PR #1, opened 2025-03-17
Stropt v4

christian-herber-nxp:
Creating the PR for the purpose of review only. No intention to merge.

### DIFF
--- a/newlib/libc/machine/riscv/strcpy.c
+++ b/newlib/libc/machine/riscv/strcpy.c
@@ -18,37 +18,29 @@ char *strcpy(char *dst, const char *src)
 
 #if !defined(PREFER_SIZE_OVER_SPEED) && !defined(__OPTIMIZE_SIZE__)
 #if !(__riscv_misaligned_slow || __riscv_misaligned_fast)
-  int misaligned = ((uintptr_t)dst | (uintptr_t)src) & (sizeof (long) - 1);
+  int misaligned = ((uintptr_t)dst | (uintptr_t)src) & (sizeof (uintptr_t) - 1);
   if (__builtin_expect(!misaligned, 1))
 #endif
     {
-      long *ldst = (long *)dst;
-      const long *lsrc = (const long *)src;
+      uintptr_t *pdst = (uintptr_t *)dst;
+      const uintptr_t *psrc = (const uintptr_t *)src;
 
-      while (!__libc_detect_null(*lsrc))
-	*ldst++ = *lsrc++;
+      while (!__libc_detect_null(*psrc))
+        *pdst++ = *psrc++;
 
-      dst = (char *)ldst;
-      src = (const char *)lsrc;
+      dst = (char *)pdst;
+      src = (const char *)psrc;
 
-      char c0 = src[0];
-      char c1 = src[1];
-      char c2 = src[2];
-      if (!(*dst++ = c0)) return dst0;
-      if (!(*dst++ = c1)) return dst0;
-      char c3 = src[3];
-      if (!(*dst++ = c2)) return dst0;
-      if (sizeof (long) == 4) goto out;
-      char c4 = src[4];
-      if (!(*dst++ = c3)) return dst0;
-      char c5 = src[5];
-      if (!(*dst++ = c4)) return dst0;
-      char c6 = src[6];
-      if (!(*dst++ = c5)) return dst0;
-      if (!(*dst++ = c6)) return dst0;
+      if (!(*dst++ = src[0])) return dst0;
+      if (!(*dst++ = src[1])) return dst0;
+      if (!(*dst++ = src[2])) return dst0;
+      if (!(*dst++ = src[3])) return dst0;
+      if (!(*dst++ = src[4])) return dst0;
+      if (!(*dst++ = src[5])) return dst0;
+      if (!(*dst++ = src[6])) return dst0;
 
 out:
-      *dst++ = 0;
+      *dst = 0;
       return dst0;
     }
 #endif /* not PREFER_SIZE_OVER_SPEED */
--- a/newlib/libc/machine/riscv/strcpy.c
+++ b/newlib/libc/machine/riscv/strcpy.c
@@ -10,49 +10,8 @@
 */
 
 #include <string.h>
-#include <stdint.h>
 
 char *strcpy(char *dst, const char *src)
 {
-  char *dst0 = dst;
-
-#if !defined(PREFER_SIZE_OVER_SPEED) && !defined(__OPTIMIZE_SIZE__)
-#if !(__riscv_misaligned_slow || __riscv_misaligned_fast)
-  int misaligned = ((uintptr_t)dst | (uintptr_t)src) & (sizeof (uintptr_t) - 1);
-  if (__builtin_expect(!misaligned, 1))
-#endif
-    {
-      uintptr_t *pdst = (uintptr_t *)dst;
-      const uintptr_t *psrc = (const uintptr_t *)src;
-
-      while (!__libc_detect_null(*psrc))
-        *pdst++ = *psrc++;
-
-      dst = (char *)pdst;
-      src = (const char *)psrc;
-
-      if (!(*dst++ = src[0])) return dst0;
-      if (!(*dst++ = src[1])) return dst0;
-      if (!(*dst++ = src[2])) return dst0;
-      if (!(*dst++ = src[3])) return dst0;
-      if (!(*dst++ = src[4])) return dst0;
-      if (!(*dst++ = src[5])) return dst0;
-      if (!(*dst++ = src[6])) return dst0;
-
-out:
-      *dst = 0;
-      return dst0;
-    }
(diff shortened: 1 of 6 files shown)


newlib/libc/machine/riscv/sys/string.h
```diff
@@ -12,12 +12,106 @@
...
 #define _SYS_STRING_H
 
+#include <stdint.h>
+
+#if __riscv_zbb
+  #include <riscv_bitmanip.h>
+
+  // Determine which intrinsics to use based on word size and endianness
+  #if __riscv_xlen == 64
+    #define __LIBC_RISCV_ZBB_ORC_B       __riscv_orc_b_64
```
christian-herber-nxp, line 22:
```suggestion
    #define __LIBC_RISCV_ZBB_ORC_B(x)       __riscv_orc_b_64(x)
```

newlib/libc/machine/riscv/sys/string.h
```diff
@@ -12,12 +12,106 @@
...
+    #if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+      #define __LIBC_RISCV_ZBB_CNT_Z     __riscv_ctz_64
+    #else
+      #define __LIBC_RISCV_ZBB_CNT_Z     __riscv_clz_64
+    #endif
+  #else
+    #define __LIBC_RISCV_ZBB_ORC_B       __riscv_orc_b_32
+
+    #if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+      #define __LIBC_RISCV_ZBB_CNT_Z     __riscv_ctz_32
```
christian-herber-nxp, line 33:
```suggestion
      #define __LIBC_RISCV_ZBB_CNT_Z (x)    __riscv_ctz_32(x)
```

newlib/libc/machine/riscv/sys/string.h
```diff
@@ -12,12 +12,106 @@
...
+    #else
+      #define __LIBC_RISCV_ZBB_CNT_Z     __riscv_clz_32
```
christian-herber-nxp, line 35:
```suggestion
      #define __LIBC_RISCV_ZBB_CNT_Z(x)     __riscv_clz_32(x)
```

newlib/libc/machine/riscv/sys/string.h
```diff
@@ -12,12 +12,106 @@
...
+  return __LIBC_RISCV_ZBB_ORC_B(w) != -1;
+#else
   unsigned long mask = 0x7f7f7f7f;
   if (sizeof (long) == 8)
     mask = ((mask << 16) << 16) | mask;
   return ~(((w & mask) + mask) | w | mask);
+#endif
+}
+
+static __inline char *__libc_strcpy(char *dst, const char *src, int rs)
```
christian-herber-nxp, line 57: add a comment to explain the parameter, potentially chose a more self-explaining name

christian-herber-nxp, line 57: bool_t from stdbool.h seems more fitting than int

newlib/libc/machine/riscv/sys/string.h
```diff
@@ -12,12 +12,106 @@
...
+  char ch;
+  do
+    {
+      ch = *src;
+      src++;
+      dst++;
+      *(dst - 1) = ch;
+    } while (ch);
+
+  return rs ? dst0 : dst - 1;
```
christian-herber-nxp, line 114: would be interesting to see what the compiler emits here, i.e. if there is a branch, and if this is converted to a conditional move if Zicond is present

christian-herber-nxp: oh, i guess this is being inlined in two versions?

newlib/libc/machine/riscv/strcpy.c
```diff
@@ -18,37 +18,29 @@ char *strcpy(char *dst, const char *src)
...
+      src = (const char *)psrc;
 
-      char c0 = src[0];
-      char c1 = src[1];
-      char c2 = src[2];
-      if (!(*dst++ = c0)) return dst0;
-      if (!(*dst++ = c1)) return dst0;
-      char c3 = src[3];
-      if (!(*dst++ = c2)) return dst0;
-      if (sizeof (long) == 4) goto out;
```
christian-herber-nxp, line 41: while it is correct to remove this, it may have been done in order for the compiler to eliminate the remaining statements. did you check if removing this changed something in the compiled assembly?

newlib/libc/machine/riscv/strcpy.c
```diff
@@ -18,37 +18,29 @@ char *strcpy(char *dst, const char *src)
...
-      if (!(*dst++ = c6)) return dst0;
+      if (!(*dst++ = src[0])) return dst0;
+      if (!(*dst++ = src[1])) return dst0;
+      if (!(*dst++ = src[2])) return dst0;
+      if (!(*dst++ = src[3])) return dst0;
+      if (!(*dst++ = src[4])) return dst0;
+      if (!(*dst++ = src[5])) return dst0;
+      if (!(*dst++ = src[6])) return dst0;
 
 out:
```
christian-herber-nxp: if you remove the goto, the out label should also be removed

newlib/libc/machine/riscv/strcpy.c
```diff
@@ -18,37 +18,29 @@ char *strcpy(char *dst, const char *src)
...
-      *dst++ = 0;
+      *dst = 0;
```
christian-herber-nxp: this change looks fine, but explicitly mention it in the commit message (best case, all atomic changes should be separate commit messages)

newlib/libc/machine/riscv/strcpy.c
```diff
@@ -18,37 +18,29 @@ char *strcpy(char *dst, const char *src)
...
-      if (!(*dst++ = c2)) return dst0;
-      if (sizeof (long) == 4) goto out;
-      char c4 = src[4];
-      if (!(*dst++ = c3)) return dst0;
-      char c5 = src[5];
-      if (!(*dst++ = c4)) return dst0;
-      char c6 = src[6];
-      if (!(*dst++ = c5)) return dst0;
-      if (!(*dst++ = c6)) return dst0;
+      if (!(*dst++ = src[0])) return dst0;
```
christian-herber-nxp: I  dislike the coding style (i.e. value assignment within an if clause). I understand it is compact, but it also is difficult to debug.
In general, you could consider making this a loop. At -O3, I would expect the compiler unrolls it. If needed, you can give the compiler a hint to unroll the loop via a pragma

newlib/libc/machine/riscv/strcpy.c
thebigclub: That is a very good point. My guess would be that it doesn't eliminate the remaining statements without the check, but I'll have to examine the disassembly.

newlib/libc/machine/riscv/strcpy.c
thebigclub: Whoops, totally forgot to remove that label.

newlib/libc/machine/riscv/sys/string.h
thebigclub: I'll update all these macro definitions.

newlib/libc/machine/riscv/sys/string.h
thebigclub: I'll update the type and either add a more descriptive name or a comment explaining it.

newlib/libc/machine/riscv/sys/string.h
thebigclub: I was looking at the disassembly to see if it would use Zicond, but I think because of the inlining it already optimizes it another way. But I'll make sure the extension is enabled when I compiled to verify it doesn't affect the optimizations.

newlib/libc/machine/riscv/strcpy.c
thebigclub: Your comment about the `sizeof()` check got me thinking that the code after the check gets removed when it's four bytes. I'm guessing they decided to manually unroll the loop. Maybe I could have two different versions based on the size. I'll have to think about this.

thebigclub:
I'll add a new type to `asm.h`.

newlib/libc/machine/riscv/sys/string.h
christian-herber-nxp: you can actually just apply the changes i suggested from the web interface. it is quite convenient

newlib/libc/machine/riscv/strcpy.c
christian-herber-nxp: yes, pretty much that. e.g. the removing of the goto and removing the label need to be in one commit. otherwise you would break the code. but changing the data types and removing the check is separate matters

newlib/libc/machine/riscv/strcpy.c
christian-herber-nxp: clearly they decided to manually unroll it. but as long as the compiler gives you optimal performance without manual optimization, i believe you should leave this decision to the compiler.

PR #2, opened 2025-03-18
Stropt v5

thebigclub:
After disassembling the binary, I see why the `sizeof()` checks existed. Without them the remaining code is included (at `-O3` it's nearly indecipherable but the object file has far more instructions without the check), so I added them back and left the code manually unrolled.

It's not really clear to me if by convention I should use double quotes vs. angle brackets to include `asm.h`. I see both in the codebase. I went with double quotes since I'm referring to the project's `asm.h` header file and not the one in the system header directories.

### DIFF
--- a/newlib/libc/machine/riscv/strlen.c
+++ b/newlib/libc/machine/riscv/strlen.c
@@ -11,6 +11,7 @@
 
 #include <string.h>
 #include <stdint.h>
+#include "sys/asm.h"
 
 size_t strlen(const char *str)
 {
@@ -21,33 +22,33 @@ size_t strlen(const char *str)
     ;
   return str - start - 1;
 #else
-  if (__builtin_expect ((uintptr_t)str & (sizeof (long) - 1), 0)) do
+  if (__builtin_expect ((uintxlen_t)str & (sizeof (uintxlen_t) - 1), 0)) do
     {
       char ch = *str;
       str++;
       if (!ch)
-	return str - start - 1;
-    } while ((uintptr_t)str & (sizeof (long) - 1));
+      return str - start - 1;
+    } while ((uintxlen_t)str & (sizeof (uintxlen_t) - 1));
 
-  unsigned long *ls = (unsigned long *)str;
-  while (!__libc_detect_null (*ls++))
+  uintxlen_t *ps = (uintxlen_t *)str;
+  while (!__libc_detect_null (*ps++))
     ;
-  asm volatile ("" : "+r"(ls)); /* prevent "optimization" */
+  asm volatile ("" : "+r"(ps)); /* prevent "optimization" */
 
-  str = (const char *)ls;
-  size_t ret = str - start, sl = sizeof (long);
+  str = (const char *)ps;
+  size_t ret = str - start, sp = sizeof (*ps);
 
-  char c0 = str[0 - sl], c1 = str[1 - sl], c2 = str[2 - sl], c3 = str[3 - sl];
-  if (c0 == 0)            return ret + 0 - sl;
-  if (c1 == 0)            return ret + 1 - sl;
-  if (c2 == 0)            return ret + 2 - sl;
-  if (sl == 4 || c3 == 0) return ret + 3 - sl;
+  char c0 = str[0 - sp], c1 = str[1 - sp], c2 = str[2 - sp], c3 = str[3 - sp];
+  if (c0 == 0)             return ret + 0 - sp;
+  if (c1 == 0)             return ret + 1 - sp;
+  if (c2 == 0)             return ret + 2 - sp;
+  if (sp == 4 || c3 == 0)  return ret + 3 - sp;
 
-  c0 = str[4 - sl], c1 = str[5 - sl], c2 = str[6 - sl], c3 = str[7 - sl];
-  if (c0 == 0)            return ret + 4 - sl;
-  if (c1 == 0)            return ret + 5 - sl;
-  if (c2 == 0)            return ret + 6 - sl;
+  c0 = str[4 - sp], c1 = str[5 - sp], c2 = str[6 - sp];
+  if (c0 == 0)             return ret + 4 - sp;
+  if (c1 == 0)             return ret + 5 - sp;
+  if (c2 == 0)             return ret + 6 - sp;
 
-  return ret + 7 - sl;
+  return ret + 7 - sp;
 #endif /* not PREFER_SIZE_OVER_SPEED */
 }
--- a/newlib/libc/machine/riscv/strlen.c
+++ b/newlib/libc/machine/riscv/strlen.c
@@ -31,24 +31,35 @@ size_t strlen(const char *str)
     } while ((uintxlen_t)str & (sizeof (uintxlen_t) - 1));
 
   uintxlen_t *ps = (uintxlen_t *)str;
-  while (!__libc_detect_null (*ps++))
+  uintxlen_t psval;
+
+  while (!__libc_detect_null ((psval = *ps++)))
     ;
   asm volatile ("" : "+r"(ps)); /* prevent "optimization" */
 
   str = (const char *)ps;
   size_t ret = str - start, sp = sizeof (*ps);
 
-  char c0 = str[0 - sp], c1 = str[1 - sp], c2 = str[2 - sp], c3 = str[3 - sp];
-  if (c0 == 0)             return ret + 0 - sp;
-  if (c1 == 0)             return ret + 1 - sp;
-  if (c2 == 0)             return ret + 2 - sp;
-  if (sp == 4 || c3 == 0)  return ret + 3 - sp;
+  #if __riscv_zbb
+    psval = ~__LIBC_RISCV_ZBB_ORC_B(psval);
+    psval = __LIBC_RISCV_ZBB_CNT_Z(psval);
 
-  c0 = str[4 - sp], c1 = str[5 - sp], c2 = str[6 - sp];
-  if (c0 == 0)             return ret + 4 - sp;
-  if (c1 == 0)             return ret + 5 - sp;
-  if (c2 == 0)             return ret + 6 - sp;
+    return ret + (psval >> 3) - sp;
+  #else
+    char c0 = str[0 - sp], c1 = str[1 - sp], c2 = str[2 - sp], c3 = str[3 - sp];
+    if (c0 == 0)    return ret + 0 - sp;
+    if (c1 == 0)    return ret + 1 - sp;
+    if (c2 == 0)    return ret + 2 - sp;
+    if (c3 == 0)    return ret + 3 - sp;
 
-  return ret + 7 - sp;
+    #if __riscv_xlen == 64
+      c0 = str[4 - sp], c1 = str[5 - sp], c2 = str[6 - sp];
+      if (c0 == 0)  return ret + 4 - sp;
+      if (c1 == 0)  return ret + 5 - sp;
+      if (c2 == 0)  return ret + 6 - sp;
+    #endif
+
+    return ret + 7 - sp;
+  #endif
 #endif /* not PREFER_SIZE_OVER_SPEED */
 }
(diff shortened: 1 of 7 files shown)


christian-herber-nxp:
@thebigclub : You could have just updated the old PR. I.e. force push to the previous branch, then the PR gets updated automatically. The advantage is that we could resolve the existing comments there to make sure nothing is missed.

christian-herber-nxp:
> It's not really clear to me if by convention I should use double quotes vs. angle brackets to include `asm.h`. I see both in the codebase. I went with double quotes since I'm referring to the project's `asm.h` header file and not the one in the system header directories.

You are correct.

newlib/libc/machine/riscv/sys/asm.h
```diff
@@ -12,6 +12,8 @@
 #ifndef _SYS_ASM_H
 #define _SYS_ASM_H
 
+#include <stdint.h>
```
christian-herber-nxp, line 15: just a comment on the commit message. I would add that we prefer the fixed width types, as this is robust against any ABI changes and it fits the datatypes of the intrinsics. This is more important than the C standard.
Also, i would actually squash this commit with the change where you change the datatype from long to uintxlen_t, then it is more clear what this is for.

newlib/libc/machine/riscv/strlen.c
```diff
@@ -21,33 +22,42 @@ size_t strlen(const char *str)
...
+
+    return ret + (psval >> 3) - sp;
+  #else
+    char c0 = str[0 - sp], c1 = str[1 - sp], c2 = str[2 - sp], c3 = str[3 - sp];
+    if (c0 == 0)             return ret + 0 - sp;
+    if (c1 == 0)             return ret + 1 - sp;
+    if (c2 == 0)             return ret + 2 - sp;
+    if (sp == 4 || c3 == 0)  return ret + 3 - sp;
+
+    c0 = str[4 - sp], c1 = str[5 - sp], c2 = str[6 - sp];
```
christian-herber-nxp, line 55: maybe we need to guard this for RV64 only? I suspect the compiler is able to optimize this out, but we should not rely on it.

newlib/libc/machine/riscv/sys/string.h
```diff
@@ -12,12 +12,115 @@
...
 #define _SYS_STRING_H
 
-static __inline unsigned long __libc_detect_null(unsigned long w)
+#include <stdbool.h>
+#include "asm.h"
+
+#if __riscv_zbb
+  #include <riscv_bitmanip.h>
+
+  // Determine which intrinsics to use based on word size and endianness
```
christian-herber-nxp, line 21: not based on word size, based on XLEN.

newlib/libc/machine/riscv/sys/string.h
```diff
@@ -12,12 +12,115 @@
...
+#if __riscv_zbb
+  /*
+    If there are any zeros in each byte of the register, all bits will
+    be unset for that byte value, otherwise, all bits will be set.
+    If the value is -1, all bits are set, meaning no null byte was found.
+  */
+  return __LIBC_RISCV_ZBB_ORC_B(w) != -1;
+#else
+  uintxlen_t mask = 0x7f7f7f7f;
+  if (sizeof (w) == 8)
```
christian-herber-nxp, line 53: you can replace the checks on sizeof long with #if __riscv_xlen == 64

newlib/libc/machine/riscv/sys/string.h
```diff
@@ -12,12 +12,115 @@
...
+
+      dst = (char *)pdst;
+      src = (const char *)psrc;
+
+      if (ret_start)
+        {
+          if (!(*dst++ = src[0])) return dst0;
+          if (!(*dst++ = src[1])) return dst0;
+          if (!(*dst++ = src[2])) return dst0;
+          if (sizeof (*pdst ) == 4) goto out;
```
christian-herber-nxp, line 84: simply guarding the additional checks with #if __riscv_xlen == 64 will lead to the same result, and you can avoid the goto

newlib/libc/machine/riscv/sys/string.h
```diff
@@ -12,12 +12,115 @@
...
 }
 
+
+static __inline char *__libc_strcpy(char *dst, const char *src, bool ret_start)
+{
+  char *dst0 = dst;
+
+#if !defined(PREFER_SIZE_OVER_SPEED) && !defined(__OPTIMIZE_SIZE__)
+#if !(__riscv_misaligned_slow || __riscv_misaligned_fast)
+  int misaligned = ((uintxlen_t)dst | (uintxlen_t)src) & (sizeof (uintxlen_t) - 1);
```
christian-herber-nxp, line 66: this comment is going a bit beyond your current optimizations. But I could imagine that it makes sense to have additional checks on alignment. E.g. if the data is 32-bit aligned on a RV64 machine, you could still be a lot faster if there was a branch for that. Same in case we are half-word aligned.

newlib/libc/machine/riscv/sys/asm.h
thebigclub: Hah, it's ironic because I thought you would've wanted the new data type to be its own commit so I made it separate. I can squash them. I'll also update the commit message.

newlib/libc/machine/riscv/strlen.c
thebigclub: I'll change it to the `#if __riscv_xlen == 64` check you mentioned elsewhere.

newlib/libc/machine/riscv/sys/string.h
thebigclub: I'll update the comment.

newlib/libc/machine/riscv/sys/string.h
thebigclub: I'll update this check.

newlib/libc/machine/riscv/sys/string.h
thebigclub: I'll make this change as well.

christian-herber-nxp:
> I'm confused about how this whole process is supposed to work. I thought the workflow is that you submit a patch series, collect feedback, update the patch series, and finally submit it again. I don't want to add commits on top of my series, since that would mess with the final patch series that I ultimately send to the upstream mailing list.
> 
> Are you saying that I force push to where the original branch started, replacing all of the old commits with new ones?

Here you are not sending out patches via email, but you have a pull request. Pull requests are updated after getting review input. You could add additional commits, but this does not lead to a nice series of commits. Therefore, just update your existing commits and add where needed, and the push -f to the existing branch (deleting the old commits and adding the new ones). 
Do not do this now, but just in future.

newlib/libc/machine/riscv/sys/string.h
christian-herber-nxp: you cannot align the pointer if the two inputs have different alignment, you can just use the common denominator. I believe whatever Mahmoud was working on would go into a similar direction.
But you do not need to worry about it immediately. Your changes are fine as they are. It could just be a nice improvement in future.

christian-herber-nxp:
when you have updated this PR, mark all the review items that are open as resolved. This way, we can track which changes have been done.

newlib/libc/machine/riscv/sys/string.h
thebigclub: Didn't make any change for this, but will keep it in mind.

thebigclub:
@christian-herber-nxp New changes are ready. I force pushed the branch.

christian-herber-nxp:
Interestingly, there is some github workflows in the newlib repository, one of which is failing. It is not failing on Marlene's PR. I assume this is due to the changes you did around adding a new file.
Please dig into this.

christian-herber-nxp:
> @christian-herber-nxp I think it's an issue with their check system. I see logs like this:
> 
> ```
> ============================
> 
> fstat02     1  FAIL  :  Functionality of fstat(2) on 'testfile' Failed
> FAIL winsup.api/ltp/fstat02.exe (exit status: 1)
> 
> FAIL: winsup.api/ltp/fstat04
> ============================
> 
> fstat01     1  FAIL  :  Functionality of fstat(2) on 'testfile' Failed
> FAIL winsup.api/ltp/fstat04.exe (exit status: 1)
> 
> XFAIL: winsup.api/ltp/setgroups01
> =================================
> 
> setgroups01    1  BROK  :  Must be ROOT to run this test.
> XFAIL winsup.api/ltp/setgroups01.exe (exit status: 2)
> 
> XFAIL: winsup.api/ltp/setuid02
> ==============================
> 
> setuid02    1  BROK  :  Must be root for this test!
> XFAIL winsup.api/ltp/setuid02.exe (exit status: 2)
> 
> FAIL: winsup.api/ltp/stat01
> ===========================
> 
> stat01      1  FAIL  :  Functionality of stat(2) on 'testfile' Failed
> FAIL winsup.api/ltp/stat01.exe (exit status: 1)
> 
> XFAIL: winsup.api/ltp/umask03
> 
> --snip--
> 
> umask03   447  FAIL  :  got mode 770 expected 2 mask 775 did not take
> umask03   448  FAIL  :  got mode 770 expected 1 mask 776 did not take
> umask03   449  FAIL  :  got mode 770 expected 0 mask 777 did not take
> XFAIL winsup.api/ltp/umask03.exe (exit status: 1)
> 
> XFAIL: winsup.api/ltp/unlink08
> ==============================
> 
> unlink08    1  FAIL  :  unlink(<unwritable directory>) returned 0, expected -1, errno:13
> unlink08    2  FAIL  :  unlink(<unsearchable directory>) returned 0, expected -1, errno:13
> unlink08    3  FAIL  :  unlink(<directory>) Failed, errno=21, expected errno:1
> XFAIL winsup.api/ltp/unlink08.exe (exit status: 1)
> 
> XFAIL: winsup.api/samples/sample-fail
> =====================================
> 
> XFAIL winsup.api/samples/sample-fail.exe (exit status: 1)
> =============================
> ```
> 
> Pretty strange that it worked fine on Fedora but not Windows.
> 
> [test-suite.log](https://github.com/user-attachments/files/19342176/test-suite.log)

the checks fail on your and mahmouds changes. you both introduced a new file. On marlenes PR it works.

thebigclub:
When I run `git ls-files -s *` in the relevant directory (snippet):
```
100644 9243457b25a20318dcf0c4f3587282641f837e56 0	stpcpy.c
100644 12c39db94c06a8fbab120145177f98590a262b5b 0	strcmp.S
100644 f770493fbc2daa30af0c9ff6481ce055186310ba 0	strcpy.c
100644 7e5d41617eac291d54f44e2a0df5d32d02692d8c 0	strlen.c
```
`stpcpy.c` is the new file I added. It has the same Git permissions as the other files.

I'm going to try running the test again and see if anything changes.

thebigclub:
@christian-herber-nxp
```
Copying 'C:\Users\runneradmin\.gitconfig' to 'D:\a\_temp\75c66f47-bc67-48d8-be82-195919f11e13\.gitconfig'
Temporarily overriding HOME='D:\a\_temp\75c66f47-bc67-48d8-be82-195919f11e13' before making global git config changes
Adding repository directory to the temporary git global config as a safe directory
C:\cygwin\bin\git.exe config --global --add safe.directory D:\a\newlib\newlib
##[debug]0
##[debug]
C:\cygwin\bin\git.exe config --local --name-only --get-regexp core\.sshCommand
fatal: --local can only be used inside a git repository
##[debug]128
##[debug]
C:\cygwin\bin\git.exe submodule foreach --recursive "sh -c \"git config --local --name-only --get-regexp 'core\.sshCommand' && git config --local --unset-all 'core.sshCommand' || :\""
fatal: detected dubious ownership in repository at '/cygdrive/d/a/newlib/newlib'
To add an exception for this directory, call:

	git config --global --add safe.directory /cygdrive/d/a/newlib/newlib
##[debug]Unsetting HOME override
Warning: The process 'C:\cygwin\bin\git.exe' failed with exit code 128
##[debug]Node Action run completed with exit code 0
##[debug]Finishing: Post Run actions/checkout@v3
```

I don't think this has any relevance to what we're doing. Seems like a GitHub issue to me. If you remember the `stropt_v4` PR, that also added a file and had no issue with the Windows build. How about we ignore this?

If you don't have any more feedback for this PR, I can write up a cover letter and send it to you. Once that's good I can email the Newlib mailing list.

christian-herber-nxp:
yes, write it up and get ready to send it out.

PR #3, opened 2025-03-18
memmove code size optimization

christian-herber-nxp:
…nstructions can be compressed

### DIFF
--- a/newlib/libc/machine/riscv/memmove.S
+++ b/newlib/libc/machine/riscv/memmove.S
@@ -14,26 +14,26 @@
 .global memmove
 .type	memmove, @function
 memmove:
-  beqz a2, 2f
+  beqz a2, .Ldone
 
   mv a4, a0
   li a3, 1
-  bgtu  a1, a0, 1f
+  bgtu  a1, a0, .Lcopy
 
   li a3, -1
   add   a4, a4, a2
   add   a1, a1, a2
 
-3:
+.Lincrement:
   add   a4, a4, a3
   add   a1, a1, a3
-1:
+.Lcopy:
   lb a5, 0(a1)
   sb a5, 0(a4)
   add   a2, a2, -1
-  bnez a2, 3b
+  bnez a2, .Lincrement
 
-2:
+.Ldone:
   ret
 
   .size	memmove, .-memmove
--- a/newlib/libc/machine/riscv/memmove.S
+++ b/newlib/libc/machine/riscv/memmove.S
@@ -28,9 +28,9 @@ memmove:
   add   a4, a4, a3 		/* in case of source address < destination address, increment both addresses by -1 before copying any data to obtain the correct start addresses */
   add   a1, a1, a3
 .Lcopy:
-  lbu a5, 0(a1)			/* copy bytes as long as a2 (= the number of bytes to be copied) > 0 */
+  lbu a5, 0(a1)
+  add   a2, a2, -1		/* copy bytes as long as a2 (= the number of bytes to be copied) > 0. the increment is done here to relax the RAW dependency between load and store */
   sb a5, 0(a4)
-  add   a2, a2, -1
   bnez a2, .Lincrement
 
 .Ldone:
--- a/newlib/libc/machine/riscv/memmove.S
+++ b/newlib/libc/machine/riscv/memmove.S
@@ -26,8 +26,8 @@ memmove:
   add a1, a1, a4
 
 1:
-  lb t2, 0(a1)
-  sb t2, 0(t1)
+  lb a5, 0(a1)
+  sb a5, 0(t1)
   add   a2, a2, -1
   add   t1, t1, a3
   add   a1, a1, a3
--- a/newlib/libc/machine/riscv/memmove.S
+++ b/newlib/libc/machine/riscv/memmove.S
@@ -14,21 +14,21 @@
 .global memmove
 .type	memmove, @function
 memmove:
-  beqz a2, .Ldone
+  beqz a2, .Ldone		/* in case there are 0 bytes to be copied, return immediately */
 
-  mv a4, a0
+  mv a4, a0			/* copy the destination address over to a4, since memmove should return that address in a0 at the end */
   li a3, 1
-  bgtu  a1, a0, .Lcopy
+  bgtu  a1, a0, .Lcopy		/* in case of source address > destination address, copy from start to end of the specified memory area */
 
-  li a3, -1
-  add   a4, a4, a2
-  add   a1, a1, a2
+  li a3, -1			/* otherwhise, we need to start copying from the end of the specified memory area, therefore after each copied byte, increment the addresses by -1 */
+  add   a4, a4, a2		/* add the number of bytes to be copied to both addresses. this gives an incorrect address, */
+  add   a1, a1, a2		/* therefore we need to subtract 1 from both addresses in the next step before starting the copying process. */
 
 .Lincrement:
-  add   a4, a4, a3
+  add   a4, a4, a3 		/* in case of source address < destination address, increment both addresses by -1 before copying any data to obtain the correct start addresses */
   add   a1, a1, a3
 .Lcopy:
-  lb a5, 0(a1)
+  lb a5, 0(a1)			/* copy bytes as long as a2 (= the number of bytes to be copied) > 0 */
   sb a5, 0(a4)
   add   a2, a2, -1
   bnez a2, .Lincrement
--- a/newlib/libc/machine/riscv/memmove.S
+++ b/newlib/libc/machine/riscv/memmove.S
@@ -16,22 +16,22 @@
 memmove:
   beqz a2, 2f
 
-  mv t1, a0
+  mv a4, a0
   li a3, 1
   bgtu  a1, a0, 1f
 
   li a3, -1
-  addi  a4, a2 , -1
-  add t1, t1, a4
-  add a1, a1, a4
+  add   a4, a4, a2
+  add   a1, a1, a2
 
+3:
+  add   a4, a4, a3
+  add   a1, a1, a3
 1:
   lb a5, 0(a1)
-  sb a5, 0(t1)
+  sb a5, 0(a4)
   add   a2, a2, -1
-  add   t1, t1, a3
-  add   a1, a1, a3
-  bnez a2, 1b
+  bnez a2, 3b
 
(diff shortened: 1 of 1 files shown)


newlib/libc/machine/riscv/memmove.S
```diff
@@ -9,31 +9,32 @@
...
 .type	memmove, @function
-memmove:
-  beqz a2, 2f
 
-  mv t1, a0
-  li a3, 1
-  bgtu  a1, a0, 1f
+#if defined(PREFER_SIZE_OVER_SPEED) || defined(__OPTIMIZE_SIZE__)
+memmove:
+  beqz a2, 3f			// if 0 bytes are to be copied, return immediately
```
christian-herber-nxp, line 19: the style of comments you are using are technically speaking C++ comments. While most modern compilers do know how to handle these also C-code, it is better practice to use /* comment */ in code compiled with a C compiler

newlib/libc/machine/riscv/memmove.S
```diff
@@ -9,31 +9,32 @@
...
 .text
 .global memmove
 .type	memmove, @function
-memmove:
-  beqz a2, 2f
 
-  mv t1, a0
-  li a3, 1
-  bgtu  a1, a0, 1f
+#if defined(PREFER_SIZE_OVER_SPEED) || defined(__OPTIMIZE_SIZE__)
```
christian-herber-nxp, line 17: why did you move this?

newlib/libc/machine/riscv/memmove.S
```diff
@@ -9,31 +9,32 @@
...
+memmove:
+  beqz a2, 3f			// if 0 bytes are to be copied, return immediately
+  mv a3, a0			// copy destination address to a3, since we want to return the destination address in a0 at the end
+  li a4, 1
+  bgtu  a1, a0, 2f
```
christian-herber-nxp, line 22: there is a lot of trailing whitespaces here. Configure your editor to show these. You might also be able to configure your editor to trim trailing whitespaces on save.

newlib/libc/machine/riscv/memmove.S
m-fally: actually this is a mistake, I will put it back!

newlib/libc/machine/riscv/memmove.S
```diff
@@ -9,31 +9,32 @@
...
+  add  a3, a3, a2		// if the destination address is smaller than the source address, we start copying from the end of the specified memory area
+  add  a1, a1, a2	
 
 1:
-  lb t2, 0(a1)
-  sb t2, 0(t1)
-  add   a2, a2, -1
-  add   t1, t1, a3
-  add   a1, a1, a3
+  add   a3, a3, a4
```
christian-herber-nxp, line 29: i keep looking at this and feeling it is incorrect. in the case where you start from the back, you are decrementing the addresses before you even moved the first byte. I believe you will miss one byte in the move process.

christian-herber-nxp: the tests passed though, so I am not sure if I am just not getting what this does

m-fally: I think the confusing thing might be, that i am first adding n (the number of bytes to be copied) to both addresses. but that gives me an incorrect address, because the "last" (which in this case becomes the first, since we are copying from back to front) byte that should be copied is the byte at (sourceaddress + n) - 1

In the original this was done by having an extra register that contained the value of n-1 and then that value was added to the addresses, but i needed to find a way to write the function with less registers, since otherwhise I would have ran out of compressed-instruction-compatible registers.

however i hope i am not overlooking something here myself...

christian-herber-nxp: got it. makes sense. i did double check the test suite and this does test the backwards overlapping scenario, so we should have some confidence in the correctness of the solution.

christian-herber-nxp: i think the optimization is good and the opportunities are exhausted. Better to focus on the performance optimized version :)

thebigclub: Unless I'm mixing something up, this comment:

`// if the destination address is smaller than the source address, we start copying from the end of the specified memory area`

Is backwards. `a0` is the destination and `a1` the source. So if this condition is false:

`bgtu  a1, a0, 2f`

The branch is not taken because the source is less than (or equal to) the destination. So the comment should say that the source address is smaller than the destination. Could also get rid of the "if" part in the beginning, since at that line you know it's true.

christian-herber-nxp: in general, comments that describe the why are more useful than ones that describe the what.
The why in this case would be that you need to prevent overwriting before moving in case there is an overlap.

newlib/libc/machine/riscv/memmove.S
```diff
@@ -9,31 +9,32 @@
...
+  add   a1, a1, a4
+2:
+  lb a5, 0(a1)
```
christian-herber-nxp, line 32: this can be lbu (this is slightly more precise, although the final behavior is identical)

m-fally:
I have gathered the following data on the improvement of code density:

### Static code size
Measuring the static code size of the object files generated with the old and new implementations gives the following results:

new implementation compiled with rv32ic: 34 bytes
old implementation compiled with rv32ic: 38 bytes

new implementation compiled with rv32ic_zcb: 30 bytes
old implementation compiled with rv32ic_zcb: 38 bytes


### Dynamic code size
Across 4096 calls to memmove, I collected the differences of the retired instruction counts before and after each call to memmove and calculated the following average values:

**Arithmetic mean:**
new implementation compiled with rv32ic: 291.522
old implementation compiled with rv32ic: 293.0139

new implementation compiled with rv32ic_zcb: 291.522
old implementation compiled with rv32ic_zcb: 293.0139


**Median**
new implementation compiled with rv32ic: 306
old implementation compiled with rv32ic: 308

new implementation compiled with rv32ic_zcb: 306
old implementation compiled with rv32ic_zcb: 308

newlib/libc/machine/riscv/memmove.S
m-fally: this is fixed now

newlib/libc/machine/riscv/memmove.S
m-fally: this is also fixed

newlib/libc/machine/riscv/memmove.S
```diff
@@ -9,31 +9,32 @@
...
-  lb t2, 0(a1)
-  sb t2, 0(t1)
-  add   a2, a2, -1
-  add   t1, t1, a3
-  add   a1, a1, a3
+  add   a3, a3, a4
+  add   a1, a1, a4
+2:
+  lb a5, 0(a1) 			
+  sb a5, 0(a3)
```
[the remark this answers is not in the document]
m-fally: this is also fixed now

newlib/libc/machine/riscv/memmove.S
m-fally: this is also fixed

m-fally:
> > Dynamic code size
> 
> small nitpick, but technically speaking, the dynamic code size would be in bytes and would have to consider also the length of each retired instruction. This is more like a dynamic instruction count. Totally sufficient, especially as you are size optimizing.
> 
> Please, for the comments that I made on you code, reply if you fixed this. I can then mark the conversation as resolved to track if there are any open items.

I see, I've updated the comment to say instruction count instead of code size!
Also I've now added replies to all your earier comments, if you want I can also resolve the conversations?

newlib/libc/machine/riscv/memmove.S
```diff
@@ -21,8 +21,8 @@ memmove:
   bgtu  a1, a0, 2f		/* in case of source address > destination address, copy from start to end of the specified memory area */
 
   li a3, -1			/* otherwhise, we need to start copying from the end of the specified memory area, therefore after each copied byte, increment the addresses by -1 */
-  add   a4, a4, a2		/* add the number of bytes to be copied to both addresses */
-  add   a1, a1, a2
+  add   a4, a4, a2		/* add the number of bytes to be copied to both addresses. this gives an incorrect address, */
+  add   a1, a1, a2		/* therefore we need to subtract 1 from both addresses in the next step before starting the copying process. */
```
christian-herber-nxp, line 25: squash this with the previous commit

newlib/libc/machine/riscv/memmove.S
```diff
@@ -14,26 +14,26 @@
...
 
-  li a3, -1
-  addi  a4, a2 , -1
-  add t1, t1, a4
-  add a1, a1, a4
+  li a3, -1			/* otherwhise, start copying from the end of the specified memory area in order to prevent data loss in case of overlapping memory areas.*/
+  add   a4, a4, a2		/* add the number of bytes to be copied to both addresses. this gives us the address one byte past the end of the memory area we want to copy, */
+  add   a1, a1, a2		/* therefore we need to subtract 1 from both addresses in the next step before starting the copying process. */
 
 1:
```
christian-herber-nxp, line 27: I would suggest giving names to the labels instead of numbers. this helps the reader understand what's happening

christian-herber-nxp:
this is looking really good so far.
I think this is close to being upstreamable.
I would suggest we have a look at all the size optimized versions and optimize, as the optimization will be very similar.
then we can submit one big patch series.

m-fally:
> this is looking really good so far. I think this is close to being upstreamable. I would suggest we have a look at all the size optimized versions and optimize, as the optimization will be very similar. then we can submit one big patch series.

Cool! Thank you for the review, I'll make the adjustments :)
And should I then focus on the other size-optimized functions and put the memmove-performance-optimized one aside for the moment? I have not written any actual code for the latter yet, so also someone else could take it on (but I don't mind continuing with it also)

christian-herber-nxp:
If you could just get an overview of which existing code size optimized functions could benefit from the same treatment, that would be perfect. We can synchronize in monday

newlib/libc/machine/riscv/memmove.S
m-fally: this is fixed now

newlib/libc/machine/riscv/memmove.S
m-fally: this is also fixed now!

PR #4, opened 2025-03-19
memcpy optimization

i3abghany:
This pull request has two commits.

1. Exclusive usage of compressed registers for the size-optimized version (now, it's fully compressed). Size reduced by 25% (from 20b to 16b).
2. A new implementation for perf memcpy in assembly. The main gain in the new version is size optimization while not losing performance:

- The new implementation is 48% smaller than the C version compiled with -O3 with GCC 14.2, and is 25% smaller when compiled with -Os.  
- In terms of performance, measured in `spike`, the new implementation is achieving roug
hly the same performance (within 0.1% on average) as the previous C (-O3) implementation.

image

image

I have been trying to include the usage of some new extensions' instructions, but couldn't find a beneficial usage and all I could think about seemed pretty contrived. I didn't want to include new instructions for the sake of it. Please let me know if you have suggestions on this front.


christian-herber-nxp:
your axes would benefit a lot from labels :D.
We need to dig into futher improving the performance. If you are just matching the C performance, it is better to stay with the C version in my opinion.
What flags did you compile the C version with?
Maybe there are some tricks to learn from the C version which could improve the performance at the cost of code size

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -14,14 +14,14 @@
...
 .type	memcpy, @function
 memcpy:
-  mv t1, a0
+  mv a3, a0
   beqz a2, 2f
 
 1:
-  lb t2, 0(a1)
-  sb t2, 0(t1)
+  lbu a4, 0(a1)
```
christian-herber-nxp, line 21: good catch. this does not really change the function, but lb would include some unnecessary sign extension operation

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -14,14 +14,14 @@
...
+  sb a4, 0(a3)
   add   a2, a2, -1
```
christian-herber-nxp, line 23: i have a preference to interleave the adds with the memory operation, in case, that there is some latency associated with the load operation, you reduce the possible stalls.

newlib/libc/machine/riscv/memcpy-asm.S
christian-herber-nxp: make the change from lb to lbu a separate commit. It is atomic, and this allows you to clearly give the rationale for the change in the commit message.

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -28,5 +28,91 @@ memcpy:
...
+    addi   a0, a0, 1
+    addi   a1, a1, 1
+    addi   a2, a2, -1
+    j      .Lwordify
+
+.Laligned:
+#ifdef __riscv_misaligned_avoid
+    andi   a3, a1, 3
+    bnez   a3, .Lcopy1
 #endif
```
christian-herber-nxp, line 49: there should be an else here

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -28,5 +28,91 @@ memcpy:
...
+
+.Lcopy32:
+    li     a3, 32
+    blt    a2, a3, .Lcopy16
+    lw     t1, 0(a1)
+    lw     t2, 4(a1)
+    lw     t3, 8(a1)
```
christian-herber-nxp, line 56: this does not work with rv32e. t3 is not contained in rv32e. If you want to have one assembly version for RISC-V, you need to work with x0-x15 only.

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -28,5 +28,91 @@ memcpy:
...
+.Laligned:
+#ifdef __riscv_misaligned_avoid
+    andi   a3, a1, 3
+    bnez   a3, .Lcopy1
 #endif
+
+.Lcopy32:
+    li     a3, 32
+    blt    a2, a3, .Lcopy16
+    lw     t1, 0(a1)
```
christian-herber-nxp, line 54: you are not exploiting ld in the case of RV64

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -28,5 +28,91 @@ memcpy:
 2:
   ret
 
-  .size	memcpy, .-memcpy
+#else
+.Lwordify:
+    andi   a3, a0, 3
+    beqz   a3, .Laligned
```
christian-herber-nxp, line 34: if i see this correctly, you are always checking for 64-bit alignment, but then you only use 32-bit memory accesses

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -28,5 +28,91 @@ memcpy:
 2:
   ret
 
-  .size	memcpy, .-memcpy
+#else
+.Lwordify:
```
christian-herber-nxp, line 32: why wordify? Looks more like align

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -28,5 +28,91 @@ memcpy:
...
+    lbu    a3, 0(a1)
+    sb     a3, 0(a0)
+    addi   a0, a0, 1
+    addi   a1, a1, 1
+    addi   a2, a2, -1
+    j      .Lwordify
+
+.Laligned:
+#ifdef __riscv_misaligned_avoid
+    andi   a3, a1, 3
```
christian-herber-nxp, line 47: add a comment here why this is beneficial

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -28,5 +28,91 @@ memcpy:
 2:
   ret
 
-  .size	memcpy, .-memcpy
+#else
+.Lwordify:
+    andi   a3, a0, 3
+    beqz   a3, .Laligned
+    li    a3, 3
```
christian-herber-nxp, line 35: add a comment why this is beneficial

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -28,5 +28,91 @@ memcpy:
...
+    bleu  a2, a3, .Lcopy1
+
+    lbu    a3, 0(a1)
```
christian-herber-nxp, line 38: i feel there is an opportunity to share the code with Lcopy1

newlib/libc/machine/riscv/memcpy-asm.S
christian-herber-nxp, line 34: as an additional improvement for performance, you can explore if going into a different alignment function based on the value of a3 could be an option. I.e. have alignment functions that use lw and or lh to achive quicker alignment.
The alternative is to use a similar approach as in memset, where you jump into a sequence of sb/lb instructions at an offset. For sure this would be better than the current solution, but I would like to also explore what i mentioned above

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -9,24 +9,110 @@
...
+    sw     a6, 12(a0)
+
+    addi   a1, a1, 16
+    addi   a0, a0, 16
+    addi   a2, a2, -16
+    j      .Lcopy16
+
+.Lcopy4:
+    li     a3, 4
+    blt    a2, a3, .Lcopy1
```
christian-herber-nxp, line 97: why are we looping here? if you are just copying the remainder, there should just be one word sized datum be left to copy. This also leads to a simplification where you do not really need to keep decrementing the remaining bytes, because it becomes a fixed sequence.

christian-herber-nxp:
as the change on the size optimized version is pretty simple, lets have separate PRs on the two. 
In fact, when submitting upstream, we should combine all of these changes into one patch.
I think there is enough room for improvement in the assembly version that we will make this faster than the C version

christian-herber-nxp:
I believe we need to discuss the algorithm in a more general sense. 
You have to deal with a number of different scenarios including
- one of src/dst is not XLEN aligned
- neither is XLEN aligned, and they have different alignment
- neither is XLEN aligned, but they have the same alignment
- both of src/dst are XLEN aligned

When in doubt, you should always prioritize the getting alignment of the src. The is because you actually depend on the result of load, whereas the store can complete in the background.

You will want a main work loop, in which you unroll as much as the registers permit you to. This will use ld/sd for RV64 and RV32-Zilsd, and lw/sw for RV32.

In order to generate alignment, you optimally use a sequence of lb/sb, lh/sh, and optionally lw/sw.
In you notion, these procedures would be .Lcopy1 - .Lcopy6. Which one you need depends on the lsbs of src.
You could use the lsbs of src to index a table from, which you load the address of .Lcopyx and jump to that.

After that, you run the main work loop.
You will need to add a remainder loop to do the remaining possible max sized loads/stores.
When done with this, you may have some remainder to clean up. based on the remaining bytes, you can again use something like  .Lcopy1 - .Lcopy6.

This should give you the optimal sequence of loads/store with the least overhead.

newlib/libc/machine/riscv/memcpy-asm.S
i3abghany: I am jumping from copy32 to copy4. so we can get into copy4 with 31 bytes remaining. Having copy32, copy16, copy8, and copy4 doesn't give any extra perf

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -28,5 +28,150 @@ memcpy:
...
   ret
 
-  .size	memcpy, .-memcpy
+#else
+    mv a3, a0
+    li    a4, 15
+    bleu  a2, a4, .Lcopy4
+
+.Lalign:
+#if __riscv_xlen == 32
```
christian-herber-nxp, line 37: you could avoid the ifdef if you replace do 
andi a4, a3, SZREG - 1 (SZREG is defined in sys/asm.h)

newlib/libc/machine/riscv/memcpy-asm.S
```diff
@@ -28,5 +28,150 @@ memcpy:
 2:
   ret
 
-  .size	memcpy, .-memcpy
+#else
+    mv a3, a0
```
christian-herber-nxp, line 32: the alignment of the operands is strange here. try and have it consistent within the file.

christian-herber-nxp:
@i3abghany in your statements on performance, I am currently missing details on how you did the build. E.g. for C, which flags did you enable. Did you use -mno-strict-align?

newlib/libc/machine/riscv/memcpy-asm.S
thebigclub: @christian-herber-nxp Your remark made me realize that `sys/asm.h` would be useful for `memset.S` as well. However, when I added the `typedef` to the header file for my string optimizations, now if I include it in a `.S` file, the assembler will choke on the `typedef`.

I was thinking about adding a preprocessor check for `__ASSEMBLER__`. I'll wait until the maintainers (hopefully) respond and see if I should include it in a `v2`, or wait until they apply the changes to the mainline branch, if it doesn't take too long.

newlib/libc/machine/riscv/memcpy.S
```diff
@@ -0,0 +1,295 @@
+/* Copyright (c) 2019  SiFive Inc. All rights reserved.
```
christian-herber-nxp, line 1: This copyrigth notice is clearly outdated. If you do significant changes, you should add yourself.

newlib/libc/machine/riscv/memcpy.S
```diff
@@ -0,0 +1,295 @@
...
+  add   a1, a1, 1
+  bnez a2, 1b
+
+2:
+  ret
+
+#else
+    mv a3, a0
+    li    a4, SZREG
+    addi  sp, sp, -2 * SZREG
```
christian-herber-nxp, line 36: there should be an option to use Zcmp

newlib/libc/machine/riscv/memcpy.S
```diff
@@ -0,0 +1,295 @@
...
+    sw    s1, 4(sp)
+#else
+    sd    s0, 0(sp)
+    sd    s1, 8(sp)
+#endif
+    bleu  a2, a4, .Lcopy1
+
+
+#if defined(__riscv_misaligned_avoid) || defined(__riscv_misaligned_slow)
+    andi a4, a1, SZREG - 1
```
christian-herber-nxp, line 48: i have a feeling these three instructions may be optimized. 
e.g. if you replace the andi, andi, sub sequence with an xor, andi SZREG - 1, i think this should work

christian-herber-nxp: actually i think an xnor will be most efficient

newlib/libc/machine/riscv/memcpy.S
```diff
@@ -0,0 +1,295 @@
...
+  sb a4, 0(a3)
+  add   a2, a2, -1
+  add   a3, a3, 1
+  add   a1, a1, 1
+  bnez a2, 1b
+
+2:
+  ret
+
+#else
```
christian-herber-nxp, line 33: in general, in your performance comparison to the C function, which exact flags did you use?

newlib/libc/machine/riscv/memcpy.S
christian-herber-nxp: actually this is debatable, as this would include the ra register. 
would an additional register be of help anywhere?
Otherwise, only opportunity is use of Zilsd for RV32

newlib/libc/machine/riscv/memcpy.S
```diff
@@ -0,0 +1,295 @@
...
+;# In the case where source and destination are not aligned, we align the
+;# destination, load SZREG bytes, one at a time, from source, and then store them
+;# to destination in one instruction.
+.Lalign_dest:
+    lbu a4, 0(a1)
+    sb  a4, 0(a3)
+    addi a1, a1, 1
+    addi a3, a3, 1
+    addi s0, s0, -1
+    addi a2, a2, -1
```
christian-herber-nxp, line 64: you are not using a2 in the loop, so don't update it inside the loop.
In compiler terms, perform loop-invariant code hoisting.

newlib/libc/machine/riscv/memcpy.S
```diff
@@ -0,0 +1,295 @@
...
+    bleu a2, s0, .Lcopy1
+
+    lbu a4, 0(a1)
+    lbu a5, 1(a1)
+    lbu s0, 2(a1)
+    lbu s1, 3(a1)
+
+#if __riscv_xlen == 32
+
+#ifdef __riscv_zbkb
```
christian-herber-nxp, line 79: @thebigclub : If multiple of you use this sort of thing, we should get this into a shared macro. 
Assuming this is the case, I would suggest that we deal with this when submitting a second patch that does this.

newlib/libc/machine/riscv/memcpy.S
christian-herber-nxp: seeing now that you use a4 later, maybe this is not a huge benefit. but at least you can move it into the path that needs it, so for the path that branches to Lalign, there could be less instructions.

newlib/libc/machine/riscv/memcpy.S
```diff
@@ -0,0 +1,295 @@
...
+    lw     a5, 52(a1)
+    lw     s0, 56(a1)
+    lw     s1, 60(a1)
+    addi   a1, a1, 64
+    sw     a4, 48(a3)
+    sw     a5, 52(a3)
+    sw     s0, 56(a3)
+    sw     s1, 60(a3)
+
+#else
```
christian-herber-nxp, line 198: you can reduce the code a lot by using the REG_S define from sys/asm.h.

newlib/libc/machine/riscv/memcpy.S
i3abghany: I'm only using O3. I vary mstrict-align and mno-strict-align with it.

newlib/libc/machine/riscv/memcpy.S
```diff
@@ -0,0 +1,295 @@
...
+    addi   a1, a1, 1
+    addi   a2, a2, -1
+    j      .Lalign
+
+.Laligned:
+.Lcopy64:
+    li     s0, 64
+    blt    a2, s0, .Lcopy32
+#if __riscv_xlen == 32
+    lw     a4, 0(a1)
```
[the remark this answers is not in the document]
i3abghany: This is what all other implementations do, and it gets generated by O3 as well. I think this pattern hides the latency of the loads better by overlapping them with stores. 1 load then 1 store would block on each store, while N independent loads can have some overlapping (i.e. N independent loads followed by N independent stores)

newlib/libc/machine/riscv/memcpy.S
```diff
@@ -0,0 +1,247 @@
...
+
+#include <sys/asm.h>
+
+
+
+#ifdef __riscv_zbkb
+
+    #if __riscv_xlen == 32
+        /* 32-bit + zbkb */
+        #define PACK4(r1, r2, r3, r4) \
```
christian-herber-nxp, line 21: as this is assembly, use a .marco/.endm definition rather than a C macro.

newlib/libc/machine/riscv/memcpy.S
```diff
@@ -0,0 +1,247 @@
...
+    PACK4(a4, a5, s0, s1)
+#if __riscv_xlen == 64
+
+    lbu t0, 4(a1)
+    lbu a5, 5(a1)
+    lbu s0, 6(a1)
+    lbu s1, 7(a1)
+
+    PACK4(t0, a5, s0, s1)
+    pack a4, a4, t0
```
christian-herber-nxp, line 108: i would move this inside the macro.
Also, I believe splat is the more fitting name. 
llvm libc has a similar library function:
https://github.com/llvm/llvm-project/blob/be6ccc98f38227db02164f17bfaf0ac86d800e4a/libc/src/string/memory_utils/op_generic.h#L129

newlib/libc/machine/riscv/memcpy.S
```diff
@@ -0,0 +1,247 @@
...
+
+    lbu a4, 0(a1)
+    lbu a5, 1(a1)
+    lbu s0, 2(a1)
+    lbu s1, 3(a1)
+
+    PACK4(a4, a5, s0, s1)
+#if __riscv_xlen == 64
+
+    lbu t0, 4(a1)
```
christian-herber-nxp, line 102: why do you need to use a different register for rv64 over rv32?

newlib/libc/machine/riscv/memcpy.S
```diff
@@ -0,0 +1,247 @@
...
+    REG_S t1, 5 * SZREG(a3)
+    REG_S t2, 6 * SZREG(a3)
+#ifndef __riscv_e
+    REG_S t3, 7 * SZREG(a3)
+    REG_S t4, 8 * SZREG(a3)
+    REG_S t5, 9 * SZREG(a3)
+    REG_S t6, 10 * SZREG(a3)
+#endif
+
+#ifdef __riscv_e
```
christian-herber-nxp, line 165: what is the performance difference between rv32e and rv32i?
It should be significant to justify this

christian-herber-nxp:
> Average over all alignment pairs:
> 
> * `-mtune=thead-c906 -mno-strict-align -O3`: 2%
> * `-mtune=thead-c906 -mstrict-align -O3`: 13%
> 
> Average over different sizes (0, 1, 2, 3, 4, 8, 16, 32, 64, 128, 200, 256, 300, 400, 512, 700, 800, 1000, 1024, 2048, 4096, 8192 16384):
> 
> * `-mtune=thead-c906 -mno-strict-align -O3`: 3%
> * `-mtune=thead-c906 -mstrict-align -O3`: 7%

I would also like to see the numbers on spike, i.e. difference in instret (you can of course take this out of gem5 also).
Also, please mention the isa string used here. Include results for rv32e/rv32i/rv64i.

PR #6, opened 2025-03-24
strcmp optimization

puranikvinit:
1. Use only compressed registers across the function.
2. Use `Zbb` extension:
    - use the `orc.b` instruction for the Hacker's Delight algorithm used to check for the presence of a null byte in a word.
    - use the `rev8` instruction for reversing the byte order in case of endianness being little endian.

`testsuite/newlib.string/strcmp-1.c` was compiled and run successfully on spike with `pk`, indicating no issues with the functionality.

Size of the size-optimized version could not be reduced, as it was pretty simple.
Size of perf-optimized version reduced from 278b to 174b.

I am currently working on getting an approximate count on the number of retired instructions, but I believe it must be lesser at the end for almost all average cases due to the reduction in the number of instructions itself.

### DIFF
--- a/newlib/libc/machine/riscv/strcmp.S
+++ b/newlib/libc/machine/riscv/strcmp.S
@@ -99,78 +99,93 @@ strcmp:
 
 .Lmismatch:
   # words don't match, but a2 has no null byte.
+  #if __riscv_zbb
+    xor a4, a2, a3
 
-#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+    #if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+      ctz a5, a4
+    #else
+      clz a5, a4
+    #endif
+    andi a5, a5, -8
 
-#if __riscv_xlen == 64
-  sll   a4, a2, 48
-  sll   a5, a3, 48
-  bne   a4, a5, .Lmismatch_upper
-  sll   a4, a2, 32
-  sll   a5, a3, 32
-  bne   a4, a5, .Lmismatch_upper
-#endif
-  sll   a4, a2, 16
-  sll   a5, a3, 16
-  bne   a4, a5, .Lmismatch_upper
+    srl a2, a2, a5
+    and a2, a2, 0xff
 
-  srl   a4, a2, 8*SZREG-16
-  srl   a5, a3, 8*SZREG-16
-  sub   a0, a4, a5
-  and   a1, a0, 0xff
-  bnez  a1, .Lfinal_upper_diff
-  ret
+    srl a3, a3, a5
+    and a3, a3, 0xff
 
-.Lmismatch_upper:
-  srl   a4, a4, 8*SZREG-16
-  srl   a5, a5, 8*SZREG-16
-  sub   a0, a4, a5
-  and   a1, a0, 0xff
-  bnez  a1, .Lfinal_upper_diff
-  ret
+    sub a0, a2, a3
+    ret
+  #else
+    #if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+      #if __riscv_xlen == 64
+        sll   a4, a2, 48
+        sll   a5, a3, 48
+        bne   a4, a5, .Lmismatch_upper
+        sll   a4, a2, 32
+        sll   a5, a3, 32
+        bne   a4, a5, .Lmismatch_upper
+      #endif
+      sll   a4, a2, 16
+      sll   a5, a3, 16
+      bne   a4, a5, .Lmismatch_upper
 
-.Lfinal_upper_diff:
-  and   a4, a4, 0xff
-  and   a5, a5, 0xff
-  sub   a0, a4, a5
-  ret
+      srl   a4, a2, 8*SZREG-16
+      srl   a5, a3, 8*SZREG-16
+      sub   a0, a4, a5
+      and   a1, a0, 0xff
+      bnez  a1, .Lfinal_upper_diff
+      ret
 
-#else
+      .Lmismatch_upper:
+        srl   a4, a4, 8*SZREG-16
+        srl   a5, a5, 8*SZREG-16
+        sub   a0, a4, a5
+        and   a1, a0, 0xff
+        bnez  a1, .Lfinal_upper_diff
+        ret
 
-#if __riscv_xlen == 64
-  srl   a4, a2, 48
-  srl   a5, a3, 48
-  bne   a4, a5, .Lmismatch_lower
-  srl   a4, a2, 32
-  srl   a5, a3, 32
-  bne   a4, a5, .Lmismatch_lower
-#endif
-  srl   a4, a2, 16
-  srl   a5, a3, 16
-  bne   a4, a5, .Lmismatch_lower
+      .Lfinal_upper_diff:
+        and   a4, a4, 0xff
+        and   a5, a5, 0xff
+        sub   a0, a4, a5
+        ret
+    #else
+      #if __riscv_xlen == 64
+        srl   a4, a2, 48
+        srl   a5, a3, 48
+        bne   a4, a5, .Lmismatch_lower
+        srl   a4, a2, 32
+        srl   a5, a3, 32
+        bne   a4, a5, .Lmismatch_lower
+      #endif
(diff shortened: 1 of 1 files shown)


newlib/libc/machine/riscv/strcmp.S
```diff
@@ -99,37 +89,8 @@ strcmp:
...
-  and   a1, a0, 0xff
-  bnez  a1, 1f
-  ret
-
-1:and   a4, a4, 0xff
-  and   a5, a5, 0xff
-  sub   a0, a4, a5
-  ret
+  rev8  a2, a2
+  rev8  a3, a3
```
puranikvinit, line 93: Had a quick question for this change:
1. Do I retain this change? Because it adds up an overhead of 2 instructions every time .Lmismatch is executed compared to the already existing code, but it is a significant size reduction.
2. If I should retain the change, should I reverse byte order for little endian or big endian?

newlib/libc/machine/riscv/strcmp.S
```diff
@@ -30,29 +30,19 @@ strcmp:
...
   .macro check_one_word i n
     REG_L a2, \i*SZREG(a0)
 
-    and   t0, a2, a5
-    or    t1, a2, a5
-    add   t0, t0, a5
-    or    t0, t0, t1
-
-    bne   t0, t2, .Lnull\i
+    orc.b a4, a2
```
christian-herber-nxp, line 43: the file still needs to compile for targets without Zbb. Check out e.g. in Erics patch how you can guard this.

newlib/libc/machine/riscv/strcmp.S
christian-herber-nxp: i like the change, but overall, it does not improve speed or code size. So it is better to keep both versions.

newlib/libc/machine/riscv/strcmp.S
```diff
@@ -30,29 +30,19 @@ strcmp:
...
-#if SZREG == 4
-  li a5, 0x7f7f7f7f
-#else
-  ld a5, mask
-#endif
-
   .macro check_one_word i n
     REG_L a2, \i*SZREG(a0)
 
-    and   t0, a2, a5
```
christian-herber-nxp, line 51: if this was a C function, you could simply call __libc_detect_null(). Eric has already optimized this for Zbb in his patch.
I do not see anything in this function that would not work in C. Maybe this is a candidate to move from asm to C

christian-herber-nxp:
I could imagine that you will find a faster algorithm for .Lmismatch using bitmanip.
E.g. if you xor the two values, all matching bytes will be all ones. If you invert, matching bytes will be all zero. You can use xnor operation for that.
Then you just count the leading zeros, and you will get the number of bits till the first mismatch. From that, calculate the byte position.

newlib/libc/machine/riscv/strcmp.S
puranikvinit: Oh yes, sorry I missed that...

christian-herber-nxp:
> I could imagine that you will find a faster algorithm for .Lmismatch using bitmanip.
> E.g. if you xor the two values, all matching bytes will be all ones. If you invert, matching bytes will be all zero. You can use xnor operation for that.
> Then you just count the leading zeros, and you will get the number of bits till the first mismatch. From that, calculate the byte position.
> 

I realized that you could even use sub instead of xnor, and then you are back to finding a NULL byte

thebigclub:
@puranikvinit If you check out the `stropt_v5` branch and open `newlib/libc/machine/riscv/sys/string.h`, you'll see examples on using the intrinsic functions and preprocessor guard macros.

newlib/libc/machine/riscv/strcmp.S
```diff
@@ -18,8 +18,8 @@ strcmp:
 #if defined(PREFER_SIZE_OVER_SPEED) || defined(__OPTIMIZE_SIZE__)
 1:
   lbu   a2, 0(a0)
-  lbu   a3, 0(a1)
   add   a0, a0, 1
+  lbu   a3, 0(a1)
```
m-fally, line 22: I guess this commit can be split into two, one for code size optimization and one for speed optimization, right? then i can add the code size optimization one to the branch that collects all the commits related to that :)

puranikvinit: Sure, will do this... I will directly commit to the branch you will be creating for all the size optimizations once it's done...

m-fally: ah okay perfect!

puranikvinit:
> I could imagine that you will find a faster algorithm for .Lmismatch using bitmanip. E.g. if you xor the two values, all matching bytes will be all ones. If you invert, matching bytes will be all zero. You can use xnor operation for that. Then you just count the leading zeros, and you will get the number of bits till the first mismatch. From that, calculate the byte position.

I tried this, works perfectly right with the `Zbb` extension. However for the fallback logic for targets without `Zbb`, we will have to implement the logic of counting the number of trailing zeroes or leading zeroes (based on the endianness of the target), which will consume a lot of instructions if we attempt to optimize it with a lookup table. And given the fact that there are already intrinsic functions for doing exactly this, I believe I will start looking in the direction of migrating to a simpler C implementation... Please correct me here if I am wrong

christian-herber-nxp:
> I tried this, works perfectly right with the `Zbb` extension. However for the fallback logic for targets without `Zbb`, we will have to implement the logic of counting the number of trailing zeroes or leading zeroes (based on the endianness of the target), which will consume a lot of instructions if we attempt to optimize it with a lookup table. And given the fact that there are already intrinsic functions for doing exactly this, I believe I will start looking in the direction of migrating to a simpler C implementation... Please correct me here if I am wrong

There is no need to change the algorithm for the non Zbb version.

puranikvinit:
> > I tried this, works perfectly right with the `Zbb` extension. However for the fallback logic for targets without `Zbb`, we will have to implement the logic of counting the number of trailing zeroes or leading zeroes (based on the endianness of the target), which will consume a lot of instructions if we attempt to optimize it with a lookup table. And given the fact that there are already intrinsic functions for doing exactly this, I believe I will start looking in the direction of migrating to a simpler C implementation... Please correct me here if I am wrong
> 
> There is no need to change the algorithm for the non Zbb version.

Wouldn't just having the same algorithm for both versions ease up things for maintainability and algorithm efficiency in general? Do you want me to try and get improvements comparing both the algorithms for the non Zbb version?

christian-herber-nxp:
> Wouldn't just having the same algorithm for both versions ease up things for maintainability and algorithm efficiency in general? Do you want me to try and get improvements comparing both the algorithms for the non Zbb version?

i cannot judge right now how different it would be. But it clearly does not make sense to count leading zeros if you do not have an instruction for that, so you will have to have both versions if you want the bitmanip acceleration

puranikvinit:
Here are some metrics I observed for `strcmp(src, dest)`:

TEST - 1:
`src` -> `(1100 0001) A B C \0`
`dest` -> `(0100 0001) A B C \0`

Cycle count with `Zbb` support on target - 30 cycles
Cycle count without `Zbb` support on target - 31 cycles

---

TEST - 2:
`src` -> `(0000 0001) A B C \0`
`dest` -> `(1000 0010) A B C \0`

Cycle count with `Zbb` support on target - 30 cycles
Cycle count without `Zbb` support on target - 31 cycles

---

TEST - 3:
`src` -> `D A B (1100 0001) \0`
`dest` -> `D A B (0100 0001) \0`

Cycle count with `Zbb` support on target - 30 cycles
Cycle count without `Zbb` support on target - 49 cycles

---

TEST - 4:
`src` -> `D A B (0000 0001) \0`
`dest` -> `D A B (1000 0010) \0`

Cycle count with `Zbb` support on target - 30 cycles
Cycle count without `Zbb` support on target - 49 cycles

---

TEST - 5 -> Random data filled in `src` and `dest`, with `MAX_BLOCK_SIZE` of 64:

Cycle count with `Zbb` support on target - 144 cycles
Cycle count without `Zbb` support on target - 413 cycles
(Above data is an average of various runs)

puranikvinit:
Code Size comparison:
with `Zbb` -> 198b
wihtout `Zbb` -> 308b

PR #7, opened 2025-03-25
memset speed optimization

thebigclub:
This is my first go at optimizing `memset()` for speed. I tried to document it as much as I could. There's a mix of `//` vs. `/**/` style comments. Not sure if this is an issue.

Overall, the speedup is dramatic for double word stores depending on the size of the input. For large copies, it takes less than half the cycles of stock Newlib. For smaller inputs, the speedup isn't nearly as much, but still at least an improvement.

Then there's the issue of store size. For smaller inputs, the double word store isn't so great for a range of inputs. It's probably a little slower for the cycle count for a small range. I need to write a script that passes an argument of the buffer size to use and automate this testing. Once that's done I'll post the results so we have a better idea.

I wanted to solicit early feedback from you so I don't end up wasting time going down the wrong path. What do you think about the jump tables? Right now the word table is quite large. The more you unroll, the greater the speedup, but there's a diminishing return. Maybe we have two jump tables for words: one for single and another for double. Depending on the buffer size, if it's on the small side, use the single word table, but if it's big enough, use double word instead. This would probably make sense only when `XLEN` is 32, though (not sure about this).

### DIFF
--- a/newlib/libc/machine/riscv/memset.S
+++ b/newlib/libc/machine/riscv/memset.S
@@ -9,105 +9,294 @@
    http://www.opensource.org/licenses.
 */
 
+#include <sys/asm.h>
+
+
+#define BYTE_TBL_SZ   31
+#define WORD_TBL_SZ   32
+
+#if __riscv_zilsd
+#undef  SZREG
+#define SZREG         8
+
+#undef  REG_S
+#define REG_S         sd
+
+/* Zilsd and Zclsd require an even numbered register */
+#define REG_SPLAT     a4
+#else
+#define REG_SPLAT     a1
+#endif
+
+/*
+  Use an extended register for Zilsd and Zclsd if available
+  since a5 is used for the odd numbered register, in order
+  to eliminate an li instruction
+*/
+#if __riscv_zilsd && !__riscv_abi_rve
+#define REG_TABLE     a6
+#else
+#define REG_TABLE     a5
+#endif
+
+
 .text
 .global memset
-.type	memset, @function
+.type  memset, @function
+
+/* void *memset(void *s, int c, size_t n); */
+
+
 memset:
 #if defined(PREFER_SIZE_OVER_SPEED) || defined(__OPTIMIZE_SIZE__)
-  mv a3, a0
-  beqz a2, .Ldone
+  mv     a3, a0
+  beqz   a2, .Ldone
 
 .Lset:
-  sb a1, 0(a3)
+  sb     a1, 0(a3)
   addi   a2, a2, -1
   addi   a3, a3, 1
-  bnez a2, .Lset
+  bnez   a2, .Lset
 
 .Ldone:
   ret
 
 #else
-  li t1, 15
-  move a4, a0
-  bleu a2, t1, .Ltiny
-  and a5, a4, 15
-  bnez a5, .Lmisaligned
+  li     REG_TABLE, BYTE_TBL_SZ
+  mv     a3, a0
 
-.Laligned:
-  bnez a1, .Lwordify
+  /* If there aren't many bytes, copy them individually to reduce overhead */
+  bleu   a2, REG_TABLE, .Lcopy_bytes
 
-.Lwordified:
-  and a3, a2, ~15
-  and a2, a2, 15
-  add a3, a3, a4
+  and    a4, a3, SZREG - 1
+  beqz   a4, .Lword_check
 
-#if __riscv_xlen == 64
-1:sd a1, 0(a4)
-  sd a1, 8(a4)
+  /*
+    Jump into the byte table depending on the number of bytes that need to be
+    written
+  */
+1:
+  auipc  t0, %pcrel_hi(.Ltable_misaligned)
+
+  /*
+    Instructions in the tables are forced to be four bytes, so scale count
+    by 4
+  */
+#if __riscv_zba
+  sh2add t0, a4, t0
 #else
-1:sw a1, 0(a4)
-  sw a1, 4(a4)
-  sw a1, 8(a4)
-  sw a1, 12(a4)
+  sll    t1, a4, 2
+  add    t0, t0, t1
 #endif
-  add a4, a4, 16
-  bltu a4, a3, 1b
 
-  bnez a2, .Ltiny
-  ret
+  /* Save the return address because we aren't exiting the function yet */
+  mv     t1, ra
+  jalr   t0, %pcrel_lo(1b)
 
-.Ltiny:
-  sub a3, t1, a2
-  sll a3, a3, 2
-1:auipc t0, %pcrel_hi(.Ltable)
-  add a3, a3, t0
+  /* Update pointer and count by what was written */
+  mv     ra, t1
+  add    a4, a4, -SZREG
+  add    a2, a2, a4
+  sub    a3, a3, a4
+
+  /* Access is now aligned. Check we can copy words. */
+  bleu   a2, REG_TABLE, .Lcopy_bytes
+
+.Lword_check:
+  /* Don't need to splat special case of zero */
+  bnez   a1, .Lsplat_byte
+#if __riscv_zilsd
+  mv     REG_SPLAT, a1
+#endif
+  j      .Lcopy_words_init
+
+/*
+  Align labels to four bytes after unconditional jumps to avoid any
+  penalties when jumping to 32-bit instructions that aren't 4-byte
+  aligned
+*/
+.p2align 2
+.Lsplat_byte:
+#if __riscv_zbkb
+  packh  REG_SPLAT, a1, a1
+#if __riscv_xlen == 64
+  packw  REG_SPLAT, REG_SPLAT, REG_SPLAT
+#endif
+  pack   REG_SPLAT, REG_SPLAT, REG_SPLAT
+#else
+  and    a1, a1, 0xFF
+  sll    t0, a1, 8
+  or     a1, a1, t0
+  sll    t0, a1, 16
+  or     REG_SPLAT, a1, t0
+#if __riscv_xlen == 64
+  sll    t0, REG_SPLAT, 32
+  or     REG_SPLAT, REG_SPLAT, t0
+#endif
+#endif
+
+.Lcopy_words_init:
+#if __riscv_zilsd
+  /* Odd register of even-odd pair */
+  mv     a5, REG_SPLAT
+#endif
+
+  /* Calculate end address */
+  and    t0, a2, ~(SZREG - 1)
+  add    t1, a3, t0
+
+  /*
+    The idea behind the table of word copies is that first we calculate any
+    remainder of bytes that need to be copied by the table that aren't an
+    entire table length. That's copied first. After that, runs of the entire
+    table are performed.
+  */
+  and    t0, t0, (WORD_TBL_SZ - 1) * SZREG
+
+  /* Skip if there's no remainder */
+  beqz   t0, .Ltable_bigly
+  neg    t0, t0
+  add    t0, t0, WORD_TBL_SZ * SZREG
+
+  /* Adjust start address with offset */
+  sub    a3, a3, t0
+
+1:
+  auipc  t2, %pcrel_hi(.Ltable_bigly)
+
+#if SZREG == 8
+  /*
+    If eight bytes are being copied with each store, we need to divide
+    the table offset in half
+  */
+  srl    t0, t0, 1
+#endif
+
+  add    t2, t2, t0
+  jr     t2, %pcrel_lo(1b)
+
+.p2align 2
+.Ltable_bigly:
+/*
+  Force the instructions to be four bytes to avoid an extra instruction
+  that would be needed to halve the offset for sw
+*/
 .option push
 .option norvc
-.Ltable_misaligned:
-  jr a3, %pcrel_lo(1b)
-.Ltable:
-  sb a1,14(a4)
-  sb a1,13(a4)
-  sb a1,12(a4)
-  sb a1,11(a4)
-  sb a1,10(a4)
-  sb a1, 9(a4)
-  sb a1, 8(a4)
-  sb a1, 7(a4)
-  sb a1, 6(a4)
-  sb a1, 5(a4)
-  sb a1, 4(a4)
-  sb a1, 3(a4)
-  sb a1, 2(a4)
-  sb a1, 1(a4)
-  sb a1, 0(a4)
+  REG_S  REG_SPLAT, SZREG*0(a3)
+  REG_S  REG_SPLAT, SZREG*1(a3)
+  REG_S  REG_SPLAT, SZREG*2(a3)
+  REG_S  REG_SPLAT, SZREG*3(a3)
+  REG_S  REG_SPLAT, SZREG*4(a3)
+  REG_S  REG_SPLAT, SZREG*5(a3)
+  REG_S  REG_SPLAT, SZREG*6(a3)
+  REG_S  REG_SPLAT, SZREG*7(a3)
+  REG_S  REG_SPLAT, SZREG*8(a3)
+  REG_S  REG_SPLAT, SZREG*9(a3)
+  REG_S  REG_SPLAT, SZREG*10(a3)
+  REG_S  REG_SPLAT, SZREG*11(a3)
+  REG_S  REG_SPLAT, SZREG*12(a3)
+  REG_S  REG_SPLAT, SZREG*13(a3)
+  REG_S  REG_SPLAT, SZREG*14(a3)
+  REG_S  REG_SPLAT, SZREG*15(a3)
+  REG_S  REG_SPLAT, SZREG*16(a3)
+  REG_S  REG_SPLAT, SZREG*17(a3)
+  REG_S  REG_SPLAT, SZREG*18(a3)
+  REG_S  REG_SPLAT, SZREG*19(a3)
+  REG_S  REG_SPLAT, SZREG*20(a3)
+  REG_S  REG_SPLAT, SZREG*21(a3)
+  REG_S  REG_SPLAT, SZREG*22(a3)
+  REG_S  REG_SPLAT, SZREG*23(a3)
+  REG_S  REG_SPLAT, SZREG*24(a3)
+  REG_S  REG_SPLAT, SZREG*25(a3)
+  REG_S  REG_SPLAT, SZREG*26(a3)
+  REG_S  REG_SPLAT, SZREG*27(a3)
+  REG_S  REG_SPLAT, SZREG*28(a3)
+  REG_S  REG_SPLAT, SZREG*29(a3)
+  REG_S  REG_SPLAT, SZREG*30(a3)
+  REG_S  REG_SPLAT, SZREG*31(a3)
 .option pop
+
+  /* Update the pointer and copy data if needed */
+  add    a3, a3, SZREG * WORD_TBL_SZ
+  bltu   a3, t1, .Ltable_bigly
+
+  /* Copy any remaining bytes */
+  and    a2, a2, SZREG - 1
+  beqz   a2, .Lexit
+
+#if __riscv_zilsd && __riscv_abi_rve
+  /* Restore table size if necessary */
+  li     REG_TABLE, BYTE_TBL_SZ
+#endif
+
+.Lcopy_bytes:
+  auipc  t0, %pcrel_hi(.Ltable_tiny)
+
+  sub    a2, REG_TABLE, a2
+
+  /*
+    Instructions in the tables are forced to be four bytes, so scale count
+    by 4
+  */
+#if __riscv_zba
+  sh2add t0, a2, t0
+#else
+  sll    a2, a2, 2
+  add    t0, t0, a2
+#endif
+
+  /* Don't save the return address because we're exiting after the jump */
+  jr     t0, %pcrel_lo(.Lcopy_bytes)
+
+.p2align 2
+.Ltable_tiny:
+/*
+  norvc is needed because the immediate is only two bits in size for c.sb,
+  and without it the table would have a mix of 2- and 4-byte instructions
+  when Zcb is available
+*/
+.option push
+.option norvc
+  sb     a1, 30(a3)
+  sb     a1, 29(a3)
+  sb     a1, 28(a3)
+  sb     a1, 27(a3)
+  sb     a1, 26(a3)
+  sb     a1, 25(a3)
+  sb     a1, 24(a3)
+  sb     a1, 23(a3)
+  sb     a1, 22(a3)
+  sb     a1, 21(a3)
+  sb     a1, 20(a3)
+  sb     a1, 19(a3)
+  sb     a1, 18(a3)
+  sb     a1, 17(a3)
+  sb     a1, 16(a3)
+  sb     a1, 15(a3)
+  sb     a1, 14(a3)
+  sb     a1, 13(a3)
+  sb     a1, 12(a3)
+  sb     a1, 11(a3)
+  sb     a1, 10(a3)
+  sb     a1, 9(a3)
+  sb     a1, 8(a3)
+#if SZREG == 8
+.Ltable_misaligned:
+#endif
+  sb     a1, 7(a3)
+  sb     a1, 6(a3)
+  sb     a1, 5(a3)
+  sb     a1, 4(a3)
+#if SZREG == 4
+.Ltable_misaligned:
+#endif
+  sb     a1, 3(a3)
+  sb     a1, 2(a3)
+  sb     a1, 1(a3)
+  sb     a1, 0(a3)
+.option pop
+.Lexit:
   ret
-
-.Lwordify:
-  and a1, a1, 0xFF
-  sll a3, a1, 8
-  or  a1, a1, a3
-  sll a3, a1, 16
-  or  a1, a1, a3
-#if __riscv_xlen == 64
-  sll a3, a1, 32
-  or  a1, a1, a3
 #endif
-  j .Lwordified
-
-.Lmisaligned:
-  sll a3, a5, 2
-1:auipc t0, %pcrel_hi(.Ltable_misaligned)
-  add a3, a3, t0
-  mv t0, ra
-  jalr a3, %pcrel_lo(1b)
-  mv ra, t0
-
-  add a5, a5, -16
-  sub a4, a4, a5
-  add a2, a2, a5
-  bleu a2, t1, .Ltiny
-  j .Laligned
-#endif
-  .size	memset, .-memset
+.size  memset, .-memset


christian-herber-nxp:
Your PR looks a bit strange now, as it includes a lot of changes.
I pushed the latest changes to the main branch here, but it did not go away. Could you double check to rebase your branch and force push it?

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,275 @@
    http://www.opensource.org/licenses.
 */
 
+#include "sys/asm.h"
+
+
+#define BYTE_TBL_SZ		15
```
christian-herber-nxp, line 15: mix of tabs and spaces

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,275 @@
...
+#define WORD_TBL_SZ   64
+
+#if __riscv_zilsd || __riscv_zclsd
```
christian-herber-nxp, line 18: zclsd implies zilsd, so can be removed

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,275 @@
...
 .global memset
 .type	memset, @function
+
+// void *memset(void *s, int c, size_t n);
+
 memset:
 #if defined(PREFER_SIZE_OVER_SPEED) || defined(__OPTIMIZE_SIZE__)
-  mv t1, a0
-  beqz a2, 2f
+  mv    t1, a0
```
christian-herber-nxp, line 33: t1 is not compressed, maybe this is an opportunity. Also tagging @m-fally

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,275 @@
...
   ret
 
 #else
-  li t1, 15
-  move a4, a0
-  bleu a2, t1, .Ltiny
-  and a5, a4, 15
-  bnez a5, .Lmisaligned
+  mv		a3, a0
+	li		t6, BYTE_TBL_SZ
```
christian-herber-nxp, line 47: make sure to not mix space and tab indention like this. It is usually a good idea to configure your editor to show the tabs and spaces explicity so you don't overlook these

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,275 @@
...
+	// If there aren't many bytes, copy them individually
```
christian-herber-nxp, line 48: // is a C++ comment. Use /* */ also for single line comments in C

christian-herber-nxp: Also, no real need to explain what is happening, but better to state why you do this. It is a cost trade-off vs. having to construct the word/doubleword to do memset

christian-herber-nxp: actually, I feel like you actually went for a different reason. this is why it should have been documented before and why you should document it now..
Just to elaborate more: Previously, I believe the limit was 15 bytes. For small vectors, it is faster to just do the byte loop, as you do not have to splat the byte. This tradeoff could actually move if you improve the splat algorithm.

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,275 @@
...
+#if __riscv_zbb
+	minu	t1, t1, a2
+#else
+	bgeu	a2, t1, 2f
+	mv		t1, a2
+#endif
+
+2:
+	/*
+		Instructions in the jump tables are forced to be four bytes, so scale
```
christian-herber-nxp, line 78: why do we force this? Originally I thought this was to ensure the jump target is 32b aligned, but I am not even sure that is the case (jumping to a 16b not 32b aligned address will likely lead to a 1 cycle penalty).

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,275 @@
...
+		count by 4
+	*/
+	sub		t2, t6, t1
+	sll		a5, t2, 2
```
christian-herber-nxp, line 82: maybe there is an opportunity for a sh2add

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,275 @@
...
+	or		a4, a1, a4
+#endif
 #endif
-  add a4, a4, 16
-  bltu a4, a3, 1b
 
-  bnez a2, .Ltiny
-  ret
+.Lcopy_words_init:
+#if __riscv_xlen == 32 && SZREG == 8
```
christian-herber-nxp, line 126: guarding with zilsd might be more explicit and easier to understand. A small comment for why this is needed helps.

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,275 @@
...
-  sub a4, a4, a5
-  add a2, a2, a5
-  bleu a2, t1, .Ltiny
-  j .Laligned
+1:
+	// Check if there are words still to be written
+	beqz	t2, .Lcopy_bytes_init
+	sub		a2, a2, t2
+
+	// Added to the pointer after the jump table
```
christian-herber-nxp, line 246: there seems to be a misconception about what a jump table is. It is not a table you jump into, but it is a table that holds addresses to jump to.

newlib/libc/machine/riscv/memset.S
m-fally: since there are not many registers used anyways i think it can simply be changed to one of the compressed registers – do you want me to make the change?

christian-herber-nxp: yes please. then eric can remove the changes that are here and just have the performance optimization in this PR.

m-fally: alright, would it make sense to make a new branch for all the code size optimizations? i could merge the memmove one into it and @i3abghany and @puranikvinit could also push their commits for memcpy and strcmp there maybe?

christian-herber-nxp: you can create this branch. start with your changes, and cherry-pick (as in git cherry-pick) the commits from the other branches. Avoid merge, you do want a clean history on that branch.

thebigclub: @christian-herber-nxp So that reference to `t1` is actually compressed. I was surprised about it and had asked in the Discord channel. @i3abghany pointed out in the documentation that you can use the full range for `mv` and it still gets compressed. We can still change it if you like.

newlib/libc/machine/riscv/memset.S
thebigclub: I'll have to look into this. Not sure what happened with the mix of tabs and spaces.

newlib/libc/machine/riscv/memset.S
thebigclub: I discovered why during disassembly. If you permit compressed instructions, the compiler dutifully converts as many instructions in the jump table to their compressed version if the immediate is small enough. You end up with a mix of compressed and uncompressed instructions, which breaks the offset used to jump into the table. They all need to be either compressed or uncompressed. Since the immediate has to be very small for `sb` (only two bits), those definitely have to be uncompressed. Might be able to get away with compressed instructions for the word table, but it depends on how large the table is. Realistically I'm going to cut down on the number of instructions after some testing, so I might be able to use compressed instructions then if they're available. Have to update the shift logic for the table in that case.

newlib/libc/machine/riscv/memset.S
thebigclub: You're right! Didn't even notice that. I'll add a check and add the instruction.

newlib/libc/machine/riscv/memset.S
thebigclub: My thought process was that if you're compiling for `rv64`, it might have `Zilsd` defined as true even though it doesn't need to use it. Is that permutation not possible? Either way I'll add a comment.

newlib/libc/machine/riscv/memset.S
thebigclub: You're right, my comment was a little sloppy. I meant the table you jump in to but the comment is saying something else. I'll fix it.

newlib/libc/machine/riscv/memset.S
thebigclub: Yeah, for small counts it wasn't worth copying words. Even with the improved splat algorithm it still was better to copy individual bytes. I'll add some comments.

newlib/libc/machine/riscv/memset.S
christian-herber-nxp: but t1 is subsequently used in lb and sb. and these cannot be compressed. if you use something else and Zcb, they could be compressed.

newlib/libc/machine/riscv/memset.S
christian-herber-nxp: it is fine to have them uncompressed. Like it stated before, if you also pay attention to enforce alignment to 32b boundary of these instructions, this benefits the performance.

newlib/libc/machine/riscv/memset.S
christian-herber-nxp: zilsd is rv32 only. impossible to have on rv64

newlib/libc/machine/riscv/memset.S
thebigclub: Ah, good point. But now I remember why I didn't change it:

```
00000000 <memset>:
   0:	86aa                	c.mv	a3,a0
   2:	c201                	c.beqz	a2,2 <memset+0x2>

00000004 <.Ltmp1>:
   4:	00b68023          	sb	a1,0(a3)
   8:	167d                	c.addi	a2,-1
   a:	0685                	c.addi	a3,1
   c:	e201                	c.bnez	a2,c <.Ltmp1+0x8>

0000000e <.Ltmp0>:
   e:	8082                	c.jr	ra
```
This is swapping `t1` with `a3`.

<br>And the original:

```
00000000 <memset>:
   0:	86aa                	c.mv	a3,a0
   2:	c201                	c.beqz	a2,2 <memset+0x2>

00000004 <.Ltmp1>:
   4:	00b68023          	sb	a1,0(a3)
   8:	167d                	c.addi	a2,-1
   a:	0685                	c.addi	a3,1
   c:	e201                	c.bnez	a2,c <.Ltmp1+0x8>

0000000e <.Ltmp0>:
   e:	8082                	c.jr	ra
```
They both end up the same size.

christian-herber-nxp: did you build with _zcb?
This is what makes the difference

thebigclub: I was missing that extension. Now I see the difference. Thank you!

newlib/libc/machine/riscv/memset.S
thebigclub: Gotcha, so if I have `zclsd` in my `-march` option, I don't need `zilsd`. Makes sense.

thebigclub: When I run `spike` with `Zclsd` but not `Zilsd`, it complains that "`'Zclsd' extension requires 'Zca' and 'Zilsd' extensions`" (I have `Zca` but not `Zilsd` specified). Is this just a peculiarity with `spike`? `clang` is fine with just `Zclsd` and I'm guessing `riscv-gnu-toolchain` will be the same as `clang` once they start supporting `Zilsd`/`Zclsd`.

thebigclub:
I think the PR looks strange because I rebased it on upstream. It includes commits from other authors.

[memset-benchmark.csv](https://github.com/user-attachments/files/19460058/memset-benchmark.csv)

Attached are my benchmark results. I have a Python script that invokes `spike` to run a binary. The C program gets passed a string length, calls `malloc()` to allocate space for that string (plus one byte for a null), and then a thousand times in a loop calls `memset()` on the same buffer. The program prints the string at the end. The Python script verifies that the string matches what it expects.

The X - Y columns in the spreadsheet compare cycle counts. If you see a negative value, it means X was faster. Positive means Y was faster. The single- and double-word versions are faster than stock Newlib for all string lengths tested (0 to 2048). Double word is faster than both with the exception of 32 values between 0 and 68 (not the entire range, just 32 values in it). I compiled all Newlib static libraries with `-O3` using `clang`.

thebigclub:
The loop unrolling for single/double words is quite large. Realistically I'll chop it down quite a bit after more testing, but I wanted to get some benchmark numbers first at least.

thebigclub:
Pushing my new changes at least got rid of the irrelevant commits from the PR, so that's helpful.

newlib/libc/machine/riscv/memset.S
christian-herber-nxp: i assume this is specific to spike, i.e. you have to explicitly list all extensions. Clang/gcc expand implied extensions

thebigclub:
Made a small optimization. Realized that if we get past the initial small table size check, don't have to worry about checking size again when aligning.

thebigclub:
Adjusted the table sizes. Made the byte table larger because it's faster than the alternative due to overhead. And shrank the single/double word table because although I'm optimizing for speed, I don't want to go crazy with using a ton of code space since these are microcontrollers we're optimizing for. My latest implementation is now faster than Newlib's current `memset()` for all counts passed in.

thebigclub:
I test the performance for a given string by passing in eight different alignments for it, so that when it's divided by 8, I get all remainders 0 through 7. That way the testing as fair as possible.

Single word copies have an advantage when it comes to alignment, since they have a remainder of 0 in two cases, vs. only one for double words, but once the count is large enough, double word copies are faster.

thebigclub:
Also fixed a bug with splatting for the 64-bit version, and shaved off an instruction for all versions when the address is already aligned.

thebigclub:
Needed to rebase and push updated branch since my latest patch was accepted.

thebigclub:
Oops, let me get that fixed.

thebigclub:
@christian-herber-nxp Thank you for linking the Linux kernel implementation. I ended up trying out their method for the table of word stores and it made that part of my algorithm faster due to fewer non-store instructions.

christian-herber-nxp:
> @christian-herber-nxp I would presume most of the time `memset()` is invoked with zero as the character to copy. I updated the code to avoid an unnecessary `mv` when only `Zilsd`/`Zclsd` need it.

that's a valid assumption

thebigclub:
Rebased to upstream and also managed to shave off a couple instructions during alignment.

thebigclub:
@christian-herber-nxp I tried to optimize with a third loop unrolling depending on the size, but it ended up being slower. I think I've squeezed out all the cycles I can, so it's ready for final review.

Assuming there are no further code changes, I can then share the detailed benchmarking results I have. Only reason I don't now is if I have to make any further code changes, I have to update the results.

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,261 @@
    http://www.opensource.org/licenses.
 */
 
+#include <sys/asm.h>
```
christian-herber-nxp, line 12: please add your copyright to this.

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,261 @@
...
   ret
 
 #else
-  li t1, 15
-  move a4, a0
-  bleu a2, t1, .Ltiny
-  and a5, a4, 15
-  bnez a5, .Lmisaligned
+  li     t6, BYTE_TBL_SZ
+  mv     a3, a0
```
christian-herber-nxp, line 54: you are using registers beyond x15. this will no longer compile with rv32e or rv64e.
Needs to be fixed.

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,261 @@
...
+#endif
 #endif
-  add a4, a4, 16
-  bltu a4, a3, 1b
 
-  bnez a2, .Ltiny
-  ret
+.Lcopy_words_init:
+#if __riscv_zilsd
+  mv     a5, REG_SPLAT
```
christian-herber-nxp, line 123: add a comment that you are moving to the odd register of an even-odd register pair

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,261 @@
...
+  srl    t0, t0, 1
+#endif
+
+  add    t2, t2, t0
+  jr     t2, %pcrel_lo(1b)
+
+.p2align 2
+.Ltable_bigly:
 .option push
 .option norvc
```
christian-herber-nxp, line 163: explain why norvc.
I think there is an opportunity to not enforce this when Zba and Zbb are enabled. Then the entire table is 16b instructions. You use half the offset in that case.
with 16b instructions, you do not even have to worry about the alignment, as they are always naturally aligned.

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,261 @@
...
+
+.Lcopy_bytes:
+  auipc  a5, %pcrel_hi(.Ltable_tiny)
+
+  sub    a2, t6, a2
+  /*
+    Instructions in the tables are forced to be four bytes, so scale count
+    by 4
+  */
+#if __riscv_zba
```
christian-herber-nxp, line 214: As this is being used a bit more often, it will make sense to create some macros for that.

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,261 @@
...
+  add    a5, a5, a2
+#endif
+
+  /* Don't save the return address because we're exiting after the jump */
+  jr     a5, %pcrel_lo(.Lcopy_bytes)
+
+.p2align 2
+.Ltable_tiny:
+.option push
+.option norvc
```
christian-herber-nxp, line 227: same comments as for the table above

christian-herber-nxp (changes requested):
Biggest change request is to bring back compliance with rv32e/rv64e.
All of riscv newlib is built this way.
This will challenge your algorithm. 
See what you can do, I am sure you will figure it out.
If there is a big benefit to having more registers, you need to either guard some parts that use additional register with the corresponding macro or create separate versions for e and i.

newlib/libc/machine/riscv/memset.S
thebigclub: Wow... that almost single-handedly sunk my implementation. I was able to salvage it with a penalty (extra `li` instruction) only to the `Zilsd`/`Zclsd` case. But it got me thinking, if a RISC-V core supports `Zilsd`/`Zclsd`, does that guarantee it will support the full 32 register set?

I'm trying to avoid the extra `li`. If I can't assume that, then would I want to check for the `__riscv_32e` macro? In other words, if a core is `rv32i`, it would not set the `__riscv_32e` macro, right? I'm trying to check for the case where `Zilsd`/`Zclsd` is available, but it's only `rv32e`.

thebigclub: Did some testing and:
```bash
$ riscv32-unknown-elf-gcc -march=rv32i -mabi=ilp32e [file]
```
Does not define the `__riscv_32e` macro. And you can have `Zilsd`/`Zclsd` with `rv32e`. But during assembling `setjmp.S` I found a bug. Newlib is assuming that you never assemble with `rv64e`, because if you do, it fails due to the registers it tries to store/load:
```bash
../libc/machine/riscv/setjmp.S: Assembler messages:
../libc/machine/riscv/setjmp.S:23: Error: illegal operands `sd s2,3*8(a0)'
../libc/machine/riscv/setjmp.S:24: Error: illegal operands `sd s3,4*8(a0)'
../libc/machine/riscv/setjmp.S:25: Error: illegal operands `sd s4,5*8(a0)'
../libc/machine/riscv/setjmp.S:26: Error: illegal operands `sd s5,6*8(a0)'
../libc/machine/riscv/setjmp.S:27: Error: illegal operands `sd s6,7*8(a0)'
../libc/machine/riscv/setjmp.S:28: Error: illegal operands `sd s7,8*8(a0)'
../libc/machine/riscv/setjmp.S:29: Error: illegal operands `sd s8,9*8(a0)'
../libc/machine/riscv/setjmp.S:30: Error: illegal operands `sd s9,10*8(a0)'
../libc/machine/riscv/setjmp.S:31: Error: illegal operands `sd s10,11*8(a0)'
../libc/machine/riscv/setjmp.S:32: Error: illegal operands `sd s11,12*8(a0)'
../libc/machine/riscv/setjmp.S:65: Error: illegal operands `ld s2,3*8(a0)'
../libc/machine/riscv/setjmp.S:66: Error: illegal operands `ld s3,4*8(a0)'
../libc/machine/riscv/setjmp.S:67: Error: illegal operands `ld s4,5*8(a0)'
../libc/machine/riscv/setjmp.S:68: Error: illegal operands `ld s5,6*8(a0)'
../libc/machine/riscv/setjmp.S:69: Error: illegal operands `ld s6,7*8(a0)'
../libc/machine/riscv/setjmp.S:70: Error: illegal operands `ld s7,8*8(a0)'
../libc/machine/riscv/setjmp.S:71: Error: illegal operands `ld s8,9*8(a0)'
../libc/machine/riscv/setjmp.S:72: Error: illegal operands `ld s9,10*8(a0)'
../libc/machine/riscv/setjmp.S:73: Error: illegal operands `ld s10,11*8(a0)'
../libc/machine/riscv/setjmp.S:74: Error: illegal operands `ld s11,12*8(a0)'
```
Do we actually need to support `rv64e`? I'm surprised no one has ever noticed this before. If I change:
```
#ifndef __riscv_32e
```
To something like:
```
#if !(defined(__riscv_32e) || defined(__riscv_64e))
```
Then it will assemble. What do you think, @christian-herber-nxp? Should we send a patch to the mailing list?

thebigclub: To add to this, I just tried rebuilding `riscv-gnu-toolchain` with `rv64e` and it fails with the same error:
```bash
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S: Assembler messages:
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:23: Error: illegal operands `sd s2,3*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:24: Error: illegal operands `sd s3,4*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:25: Error: illegal operands `sd s4,5*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:26: Error: illegal operands `sd s5,6*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:27: Error: illegal operands `sd s6,7*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:28: Error: illegal operands `sd s7,8*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:29: Error: illegal operands `sd s8,9*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:30: Error: illegal operands `sd s9,10*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:31: Error: illegal operands `sd s10,11*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:32: Error: illegal operands `sd s11,12*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:65: Error: illegal operands `ld s2,3*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:66: Error: illegal operands `ld s3,4*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:67: Error: illegal operands `ld s4,5*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:68: Error: illegal operands `ld s5,6*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:69: Error: illegal operands `ld s6,7*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:70: Error: illegal operands `ld s7,8*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:71: Error: illegal operands `ld s8,9*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:72: Error: illegal operands `ld s9,10*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:73: Error: illegal operands `ld s10,11*8(a0)'
riscv-gnu-toolchain/build/../newlib/newlib/libc/machine/riscv/setjmp.S:74: Error: illegal operands `ld s11,12*8(a0)'
make[4]: *** [Makefile:18999: libc/machine/riscv/libc_a-setjmp.o] Error 1
```
I don't think anyone has ever tried to build Newlib or the toolchain with `rv64e`.

thebigclub: `newlib/libgloss/riscv/internal_syscall.h` needs a patch as well:
```c
#ifdef __riscv_32e
  register long syscall_id asm("t0") = n;
#else
  register long syscall_id asm("a7") = n;
#endif
```

newlib/libc/machine/riscv/memset.S
thebigclub: This won't work for the byte table. The immediate for `c.sb` is only two bits in size, so a range of 0-3. You end up with a mix of compressed and uncompressed instructions:
```
...
 11c:	00b68323          	sb	a1,6(a3)
 120:	00b682a3          	sb	a1,5(a3)
 124:	00b68223          	sb	a1,4(a3)

00000128 <.Ltable_misaligned>:
 128:	8aec                	c.sb	a1,3(a3)
 12a:	8aac                	c.sb	a1,2(a3)
 12c:	8acc                	c.sb	a1,1(a3)
 12e:	8a8c                	c.sb	a1,0(a3)
```

newlib/libc/machine/riscv/memset.S
thebigclub: The reason is for speed. You're correct that all the store instructions in the word table can be compressed (unlike the byte table). But the problem then is it incurs an extra instruction to halve the offset. Since we're optimizing for speed, I want to avoid adding an instruction for the `sw` case (it's unavoidable for the `sd` case).

I've added a comment explaining this.

newlib/libc/machine/riscv/memset.S
thebigclub: It has a `#if` though. I don't think you can use preprocessor directives inside a macro. It didn't work when I tested it.

This link says it's not possible: https://stackoverflow.com/questions/301711/is-it-possible-for-c-preprocessor-macros-to-contain-preprocessor-directives/301875#301875

newlib/libc/machine/riscv/memset.S
thebigclub: Done.

newlib/libc/machine/riscv/memset.S
```diff
@@ -9,105 +9,261 @@
...
+
+.Lword_check:
+  /* Don't need to splat special case of zero */
+  bnez   a1, .Lsplat_byte
+#if __riscv_zilsd
+  mv     REG_SPLAT, a1
+#endif
+  j      .Lcopy_words_init
+
+.p2align 2
```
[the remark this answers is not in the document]
thebigclub: Done.

newlib/libc/machine/riscv/memset.S
christian-herber-nxp: > Do we actually need to support `rv64e`? I'm surprised no one has ever noticed this before. If I change:

Interesting question. It seems not. 
As I am not aware of a single device using rv64e, I would not make a big fuzz about it.

newlib/libc/machine/riscv/memset.S
christian-herber-nxp: got it.

newlib/libc/machine/riscv/memset.S
christian-herber-nxp: ok

christian-herber-nxp:
> So I changed the macro check to be more robust for this scenario. Is the first `clang` case considered a bug? Maybe I could submit an issue to their GitHub.

Clearly a bug. Would be good to file

christian-herber-nxp:
Also, no need to adjust your code to a llvm bug.

thebigclub:
@christian-herber-nxp I filed a bug report this morning and they had it fixed before lunch. Pretty nice!

Since it's addressed I updated the macros.

christian-herber-nxp:
> @christian-herber-nxp I filed a bug report this morning and they had it fixed before lunch. Pretty nice!
> 
> Since it's been addressed I updated the macros.

i see great job

PR #8, opened 2025-03-27
Code size optimizations

m-fally:
This is the branch that collects all code size optimizations we have done so far.

### DIFF
--- a/newlib/libc/machine/riscv/strcmp.S
+++ b/newlib/libc/machine/riscv/strcmp.S
@@ -18,8 +18,8 @@ strcmp:
 #if defined(PREFER_SIZE_OVER_SPEED) || defined(__OPTIMIZE_SIZE__)
 1:
   lbu   a2, 0(a0)
-  lbu   a3, 0(a1)
   addi   a0, a0, 1
+  lbu   a3, 0(a1)
   addi   a1, a1, 1
   bne   a2, a3, 2f
   bnez  a2, 1b


m-fally:
@christian-herber-nxp i noticed that in the original implementations there are some `add` instructions which should be `addi`. the assembler turns them into `addi` anyways, so I was wondering if they should be changed or if it doesn't matter?

christian-herber-nxp:
> @christian-herber-nxp i noticed that in the original implementations there are some `add` instructions which should be `addi`. the assembler turns them into `addi` anyways, so I was wondering if they should be changed or if it doesn't matter?

interesting. it seems, the assembler implements a pseudo instruction not included in the official assembly manual. i am pretty neutral on changing this.

m-fally:
> > @christian-herber-nxp i noticed that in the original implementations there are some `add` instructions which should be `addi`. the assembler turns them into `addi` anyways, so I was wondering if they should be changed or if it doesn't matter?
>
> interesting. it seems, the assembler implements a pseudo instruction not included in the official assembly manual. i am pretty neutral on changing this.

okay, then I suggest to leave it for now. and if we want we can make a last commit where we change all the affected instructions in all files once we have all the size optimized functions ready?

m-fally:
I have rebased this branch against upstream to get the strcmp()-related changes, but I'm a bit confused about the three commits from the upstream branch showing up in this PR.
I could try to make a fresh branch and cherry-pick all the commits again, but I'm not sure that's what we want?

christian-herber-nxp:
> I have rebased this branch against upstream to get the strcmp()-related changes, but I'm a bit confused about the three commits from the upstream branch showing up in this PR.
> I could try to make a fresh branch and cherry-pick all the commits again, but I'm not sure that's what we want?

it happens if the commits are not on the master branch you are merging against. need to push all these commits to it and they will be gone

christian-herber-nxp:
i like the commit message title for memove and memset. Please run a rebase updating the other commits to have the title consistent.

m-fally:
> i like the commit message title for memove and memset. Please run a rebase updating the other commits to have the title consistent.

done!

christian-herber-nxp:
Nice. Let's get the log perfect. 
Shorten the title to beprwferably 50, at most 72 chars.

You will have to drive the size optimized version thing from the title I assume.

christian-herber-nxp:
looks great. last thing to do is write the cover letter and practice sending out by via git email to us.

thebigclub:
@m-fally Check out the [`--dry-run`](https://git-scm.com/docs/git-send-email#Documentation/git-send-email.txt---dry-run) option of `git send-email`. Very useful to test before any emails actually get sent out (even when testing with us). I always use it before sending after learning the hard way, and it's helped me avoid mistakes.

m-fally:
> @m-fally Check out the [`--dry-run`](https://git-scm.com/docs/git-send-email#Documentation/git-send-email.txt---dry-run) option of `git send-email`. Very useful to test before any emails actually get sent out (even when testing with us). I always use it before sending after learning the hard way, and it's helped me avoid mistakes.

oh that's a great tip, thank you!!

puranikvinit:
Also, I have to rebase the commit appropriately before you draft another patch, will do that ASAP. Thanks!

christian-herber-nxp:
> I'm writing the commit message for the strcmp commit and I have a quick question about interleaving the instructions – since there is no dependency between the load instructions, but there is a dependency between the loads and the adds, what is the benefit of doing this?

what is the dependency with the loads and the add?
The RAW dependencies are from the loads to the branch instructions.

There is a structural conflict between the two loads, i.e. if the LSU is busy, it might not be able to start the second load. This is microarchitecture specific. Some cores are able to have multiple outstanding loads.
Long story short, it is not easy to say which is better. 
It could be interesting to investigate this with different pipeline models with gem5, to see if one or the other is better.
I can see that Vinit's solution might be better on average, as after the first access, you would expect cache hits.

As this is a performance change, I would suggest the following:
Let's submit v2 without any strcmp changes. 

Vinit can investigate the instruction order for the size optimized version with gem5.

m-fally:
I see, thank you both!
the dependency between the loads and the adds is write-after-read, unless i'm mixing something up. because e.g. the load instruction reads the address from a0 and then the addi instruction increments a0 by 1.

Then I'll prepare a version 2 now and send it out to all of you?

christian-herber-nxp:
> the dependency between the loads and the adds is write-after-read, unless i'm mixing something up. because e.g. the load instruction reads the address from a0 and then the addi instruction increments a0 by 1.

Well, this would be in the next iteration of the loop. Before that, you first have to go through the branch, which needs the loaded value.

> Then I'll prepare a version 2 now and send it out to all of you?

yes please

m-fally:
> > the dependency between the loads and the adds is write-after-read, unless i'm mixing something up. because e.g. the load instruction reads the address from a0 and then the addi instruction increments a0 by 1.
> 
> Well, this would be in the next iteration of the loop. Before that, you first have to go through the branch, which needs the loaded value.

but if the instructions are interleaved, eg. if there is a sequence like this:
```
lbu   a2, 0(a0)
addi   a0, a0, 1
```
shouldn't that be a WAR hazard no matter the iteration? sorry if i'm mixing things up!

christian-herber-nxp:
> shouldn't that be a WAR hazard no matter the iteration? sorry if i'm mixing things up!

You are right. There is a WAR, but it might not be an issue.
addi will not be blocked from execution. Once the lbu has been issued, a0 has been read and fed to the LSU. 
In an in order machine, lbu anyway has to complete before the addi result (i.e. writing to a0) is commited. The arithmetic operation of addition however can be performed before.

It will likely depend on the actual machine how this behaves, so it would be good to have a deeper investigation here. This might be really insightful.

RAW hazards are a bit more obvious, in a sense that you cannot start the execution because you do not have your input values.

m-fally:
> > shouldn't that be a WAR hazard no matter the iteration? sorry if i'm mixing things up!
> 
> You are right. There is a WAR, but it might not be an issue. addi will not be blocked from execution. Once the lbu has been issued, a0 has been read and fed to the LSU. In an in order machine, lbu anyway has to complete before the addi result (i.e. writing to a0) is commited. The arithmetic operation of addition however can be performed before.
> 
> It will likely depend on the actual machine how this behaves, so it would be good to have a deeper investigation here. This might be really insightful.
> 
> RAW hazards are a bit more obvious, in a sense that you cannot start the execution because you do not have your input values.

I see, thanks a lot!!

PR #9, opened 2025-04-11
strcmp optimization - v2

puranikvinit:
This is v2 of the strcmp optimization PR previously raised. The commits are atomic and sanitized.

I am currently working on getting stats for instruction reordering as discussed [here](https://github.com/puranikvinit/newlib/pull/8#issuecomment-2791964540). I will share the results in this PR, and if good enough, I will add the instruction interleave commit here as well.

### DIFF
--- a/newlib/libc/machine/riscv/strcmp.S
+++ b/newlib/libc/machine/riscv/strcmp.S
@@ -48,12 +48,12 @@ strcmp:
     REG_L a2, \i*SZREG(a0)
     REG_L a3, \i*SZREG(a1)
 
-    and   t0, a2, a5
+    and   a4, a2, a5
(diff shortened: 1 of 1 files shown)


christian-herber-nxp:
please update the old PR instead of creating a new one. This way I can see the history of comments etc.
You can do so by force pushing to the old branch

puranikvinit:
> please update the old PR instead of creating a new one. This way I can see the history of comments etc. You can do so by force pushing to the old branch

Okay... I thought doing so would break the PR references because the commits would be reset, but seems like that doesn't happen... I have force pushed these commits into the other branch, I will close this PR and destroy this branch then.

Thanks!

PR #10, opened 2025-04-14
setjmp optimization

puranikvinit:
There wasn't much here to optimize, I just made use of the `Zilsd` extension to reduce the code size. The entire code is basically register load/store instructions, no logic involved as such. Eric had suggested to have a look at the `Zcmp` extension, but since we are using a buffer to save the register state and not the stack itself, I don't think it'd be useful.

I got a code size reduction of `40b` when the target supports `Zilsd` extension.

### DIFF
--- a/newlib/libc/machine/riscv/setjmp.S
+++ b/newlib/libc/machine/riscv/setjmp.S
@@ -16,21 +16,33 @@
   .type   setjmp, @function
 setjmp:
 	REG_S ra,  0*SZREG(a0)
-	REG_S s0,  1*SZREG(a0)
-	REG_S s1,  2*SZREG(a0)
+  #if __riscv_xlen == 32 && (__riscv_zilsd || __riscv_zclsd)
+    sd    s0,  1*SZREG(a0)
+  #else
+	  REG_S s0,  1*SZREG(a0)
+	  REG_S s1,  2*SZREG(a0)
+  #endif
 
 #ifndef __riscv_32e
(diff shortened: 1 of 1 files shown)


thebigclub:
@puranikvinit What if you temporarily point the SP to the buffer so that the `Zcmp` instructions will then read/write to the buffer instead of the stack, and restore the SP at the end? Could that work when optimizing for space?

christian-herber-nxp:
> @puranikvinit What if you temporarily point the SP to the buffer so that the `Zcmp` instructions will then read/write to the buffer instead of the stack, and restore the SP at the end? Could that work when optimizing for space?

does not sound like something one should do. This will fall apart in case of a trap, which would rely on a proper sp value.

newlib/libc/machine/riscv/setjmp.S
```diff
@@ -16,21 +16,33 @@
   .type   setjmp, @function
 setjmp:
 	REG_S ra,  0*SZREG(a0)
-	REG_S s0,  1*SZREG(a0)
-	REG_S s1,  2*SZREG(a0)
+  #if __riscv_xlen == 32 && (__riscv_zilsd || __riscv_zclsd)
```
thebigclub, line 19: Christian [mentioned](https://github.com/puranikvinit/newlib/pull/7#discussion_r2011633187) that `Zclsd` implies `Zilsd`, so you don't need the check for `Zclsd`.

PR #11, opened 2025-04-14
RISC-V: memcpy() align dest when misaligned access is prohibited

i3abghany:
Add a code path for when source and dest are differently aligned.

If misaligned access is slow or prohibited, and the alignments of the source and destination are different, we align the destination to do word stores. This uses only one aligned store for every four (or eight for XLEN == 64) bytes of data.

This only affects the case where source and destination are differently aligned. Previously, this case was handled by copying one byte at a time. The new approach significantly reduces the number of executed (dynamic) instructions:

image

image


newlib/libc/machine/riscv/memcpy.c
```diff
@@ -58,33 +114,33 @@ memcpy(void *__restrict aa, const void *__restrict bb, size_t n)
...
 
   while (la < lend)
     BODY (la, lb, long);
 
-  a = (char *)la;
-  b = (const char *)lb;
+  a = (unsigned char *)la;
+  b = (const unsigned char *)lb;
   if (unlikely (a < end))
     goto small;
```
christian-herber-nxp, line 145: I am always in favor of getting rid of gotos if possible.
I suspect replacing this with a function call will not change the  output, but it makes debugging etc a lot nicer.

newlib/libc/machine/riscv/memcpy.c
```diff
@@ -10,46 +11,101 @@
...
-#include <string.h>
-#include <stdint.h>
 #include "../../string/local.h"
+#include <stdint.h>
+#include <string.h>
 
 #define unlikely(X) __builtin_expect (!!(X), 0)
 
+static inline long
+__load_word (const unsigned char *p)
```
christian-herber-nxp, line 24: this function should be put into the ifdef for slow misaligned access, otherwise you will have an unused function for fast access

christian-herber-nxp, line 24: this is an xlen load.
if you had rv64 with abi ilp32, you would still want this to work on 64b data.
In general, you should get rid of all cases of long and sizeof(long).
Use https://github.com/puranikvinit/newlib/blob/main/newlib/libc/machine/riscv/xlenint.h
Instead of sizeof(long), use SZREG from sys/asm.h

newlib/libc/machine/riscv/memcpy.c
```diff
@@ -10,46 +11,101 @@
...
 void *
 __inhibit_loop_to_libcall
-memcpy(void *__restrict aa, const void *__restrict bb, size_t n)
+memcpy (void *__restrict aa, const void *__restrict bb, size_t n)
 {
-  #define BODY(a, b, t) { \
-    t tt = *b; \
-    a++, b++; \
-    *(a - 1) = tt; \
+#define BODY(a, b, t)                                                         \
```
christian-herber-nxp, line 46: could this macro be replaced with an inline function?

christian-herber-nxp (changes requested):
small suggestion in comments.

i3abghany:
@christian-herber-nxp I think this looks good, so please let me know if you have other suggestions and if we can send a patch series soon.

newlib/libc/machine/riscv/memcpy.c
```diff
@@ -10,83 +11,139 @@
...
-//memcpy defined in memcpy-asm.S
+// memcpy defined in memcpy-asm.S
 #else
 
-#include <string.h>
-#include <stdint.h>
 #include "../../string/local.h"
+#include "xlenint.h"
+#include <stdint.h>
+#include <stdio.h>
```
christian-herber-nxp, line 20: what is this used for?

i3abghany: oh... that shouldn't have been there. I was using it for debugging

PR #12, opened 2025-04-27
memchr and memrchr

thebigclub:
@christian-herber-nxp Here's my initial draft of `memchr()` and `memrchr()`. I implemented them in C, based off the Newlib stock versions.

`memchr()` is very similar to `strlen()`. `memrchr()` is different enough from `memchr()` that it warrants a separate source file. Sharing the code and adding add a boolean to distinguish between the two ended up making them hard to follow and slows down the performance of both. They're different enough to where it makes sense to separate them completely.

It'd be nice to add `Zilsd`/`Zclsd` to the `rv32` version, but I don't know how feasible this is. I thought I'd bring it up to you to get your initial thoughts before any attempts.

### DIFF
--- a/newlib/libc/machine/riscv/memchr.c
+++ b/newlib/libc/machine/riscv/memchr.c
@@ -1,0 +1,99 @@
+/*
+FUNCTION
+	<<memchr>>---find character in memory
+
+INDEX
+	memchr
+
+SYNOPSIS
+	#include <string.h>
+	void *memchr(const void *<[src]>, int <[c]>, size_t <[length]>);
+
+DESCRIPTION
+	This function searches memory starting at <<*<[src]>>> for the
+	character <[c]>.  The search only ends with the first
+	occurrence of <[c]>, or after <[length]> characters; in
+	particular, <<NUL>> does not terminate the search.
+
+RETURNS
+	If the character <[c]> is found within <[length]> characters
+	of <<*<[src]>>>, a pointer to the character is returned. If
+	<[c]> is not found, then <<NULL>> is returned.
+
+PORTABILITY
+<<memchr>> is ANSI C.
+
+<<memchr>> requires no supporting OS subroutines.
+
+QUICKREF
+	memchr ansi pure
+*/
+
+#include <_ansi.h>
+#include <string.h>
+#include <limits.h>
+#include "../../string/local.h"
+
+void *
+memchr (const void *src_void,
+	int c,
+	size_t length)
+{
+  const unsigned char *src = (const unsigned char *) src_void;
+  unsigned char d = c;
+
+#if !defined(PREFER_SIZE_OVER_SPEED) && !defined(__OPTIMIZE_SIZE__)
+  unsigned long *asrc;
+  unsigned long  mask;
+  unsigned int i;
+
+  while (UNALIGNED_X(src))
+    {
+      if (!length--)
+        return NULL;
+      if (*src == d)
+        return (void *) src;
+      src++;
+    }
+
+  if (!TOO_SMALL_LITTLE_BLOCK(length))
+    {
+      /* If we get this far, we know that length is large and src is
+         word-aligned. */
+      /* The fast code reads the source one word at a time and only
+         performs the bytewise search on word-sized segments if they
+         contain the search character, which is detected by XORing
+         the word-sized segment with a word-sized block of the search
+         character and then detecting for the presence of NUL in the
+         result.  */
+      asrc = (unsigned long *) src;
+      mask = d << 8 | d;
+      mask = mask << 16 | mask;
+      for (i = 32; i < sizeof(mask) * 8; i <<= 1)
+        mask = (mask << i) | mask;
+
+      while (!TOO_SMALL_LITTLE_BLOCK(length))
+        {
+          if (DETECT_CHAR(*asrc, mask))
+            break;
+          length -= LITTLE_BLOCK_SIZE;
+          asrc++;
+        }
+
+      /* If there are fewer than LITTLE_BLOCK_SIZE characters left,
+         then we resort to the bytewise loop.  */
+
+      src = (unsigned char *) asrc;
+    }
+
+#endif /* not PREFER_SIZE_OVER_SPEED */
+
+  while (length--)
+    {
+      if (*src == d)
+        return (void *) src;
+      src++;
+    }
+
+  return NULL;
+}
--- a/newlib/libc/machine/riscv/memchr.c
+++ b/newlib/libc/machine/riscv/memchr.c
@@ -29,10 +29,15 @@ QUICKREF
 	memchr ansi pure
 */
 
-#include <_ansi.h>
-#include <string.h>
-#include <limits.h>
-#include "../../string/local.h"
+#include <sys/asm.h>
+#include <stddef.h>
+#include "rv_string.h"
+
+#if __riscv_zilsd
+#undef  SZREG
+#define SZREG           8
+#endif
+
 
 void *
 memchr (const void *src_void,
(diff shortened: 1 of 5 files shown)


thebigclub:
I have my doubts that using `Zilsd`/`Zclsd` is possible. I tried using register variables but encountered [this bug](https://github.com/llvm/llvm-project/issues/109778) in `clang`. They have a RISC-V example, and someone mentioned the `-ffixed-reg` workaround, but this is a compiler option.

thebigclub:
I used inline assembly and in theory it could work, assuming the compiler doesn't clobber the registers you choose. The binary had the correct behavior in `spike`. You have to use global register variables. I'm not really sure if you can trust both `clang` and `gcc` not to overwrite them. I can't test `gcc` since it doesn't support `Zilsd`/`Zclsd`, and I'm not sure what comprehensive testing would look like for `clang`. Relying on the compilers to not clobber the registers seems like a fragile solution.

But the real deal-breaker are the compile options. `-ffixed-x14` (`a4`) and `-ffixed-x15` (`a5`) would have to be specified for only the `memchr.c` and `memrchr.c` source files, and I don't know of any practical way of doing this in Newlib.

`Zilsd` and `Zclsd` really need intrinsic functions provided by the compiler to use them properly in `C` code.

christian-herber-nxp:
> I used inline assembly and in theory it could work, assuming the compiler doesn't clobber the registers you choose. The binary had the correct behavior in `spike`. You have to use global register variables. I'm not really sure if you can trust both `clang` and `gcc` not to overwrite them. I can't test `gcc` since it doesn't support `Zilsd`/`Zclsd`, and I'm not sure what comprehensive testing would look like for `clang`. Relying on the compilers to not clobber the registers seems like a fragile solution.
> 
> But the real deal-breaker are the compile options. `-ffixed-x14` (`a4`) and `-ffixed-x15` (`a5`) would have to be specified for only the `memchr.c` and `memrchr.c` source files, and I don't know of any practical way of doing this in Newlib.
> 
> `Zilsd` and `Zclsd` really need intrinsic functions provided by the compiler to use them properly in `C` code.

what did you inline assembly look like. The --fixed-x... options are too restrictive.
Did you use the clobbers specification? https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Clobbers-and-Scratch-Registers

However, this reminds me of the following change on llvm: https://github.com/llvm/llvm-project/commit/4615cc38f35d
I think this is what you are really looking for.

thebigclub:
So that's interesting. I didn't see the clobbers parameter. My concern is this warning about [local register variables](https://gcc.gnu.org/onlinedocs/gcc/Local-Register-Variables.html):

> Defining a register variable does not reserve the register. Other than when invoking the Extended asm, the contents of the specified register are not guaranteed. For this reason, the following uses are explicitly not supported. If they appear to work, it is only happenstance, and may stop working as intended due to (seemingly) unrelated changes in surrounding code, or even minor changes in the optimization of a future version of gcc:
>
> * Passing parameters to or from Basic asm
> * Passing parameters to or from Extended asm without using input or output operands.
> * Passing parameters to or from routines written in assembler (or other languages) using non-standard calling conventions.
>
> Some developers use Local Register Variables in an attempt to improve gcc’s allocation of registers, especially in large functions. In this case the register name is essentially a hint to the register allocator. While in some instances this can generate better code, improvements are subject to the whims of the allocator/optimizers. Since there are no guarantees that your improvements won’t be lost, this usage of Local Register Variables is discouraged.

I need to be able to refer to the values of the register pair, and if I can't use global register variables, the only choice then is local. It sounds like it might work with the `"R"` constraint. I see it [exists](https://gcc.gnu.org/onlinedocs/gcc/Machine-Constraints.html) for GCC. It's not clear to me how I would tell the compiler to use a specific variable for the odd-numbered register without local register variables, since you don't specify that register in the instruction. Maybe it would go in the clobbers list? I'll have to play around with this.

christian-herber-nxp:
> It's not clear to me how I would tell the compiler to use a specific variable for the odd-numbered register without local register variables, since you don't specify that register in the instruction. Maybe it would go in the clobbers list? I'll have to play around with this.

When specifying an even register with R, the next odd register will be picked. In this case, I would not think that you need to put it into clobbers.
This is all fairly new, so there is not a lot of prior art, but I feel like it should be somewhat straight forward

thebigclub:
It's almost working. I start with this code:

```c
          register uintxlen_t val1 asm ("a4");
          register uintxlen_t val2 asm ("a5");

          asm volatile (STRINGIFY(REG_L) " %0, 0(%1)"
                        : "=R" (val1)
                        : "r" (src)
                        : "a5"
          );
```

It disassembles to:

```bash
   11ec4: 08b5f8b3     	packh	a7, a1, a1
   11ec8: 0918c8b3     	pack	a7, a7, a7
   11ecc: 6118         	c.ld	a4, 0x0(a0)
   11ece: 01174733     	xor	a4, a4, a7
   11ed2: 28775713     	orc.b	a4, a4
   11ed6: 9f75         	c.not	a4
   11ed8: e70d         	c.bnez	a4, 0x11f02 <memchr+0xc4>
   11eda: 0116c6b3     	xor	a3, a3, a7
   11ede: 2876d713     	orc.b	a4, a3
   11ee2: 9f75         	c.not	a4
   11ee4: e705         	c.bnez	a4, 0x11f0c <memchr+0xce>
```

It's using `a3` instead of `a5`. When I get rid of the clobbers parameter:

```bash
   11ec4: 08b5f7b3     	packh	a5, a1, a1
   11ec8: 08f7c7b3     	pack	a5, a5, a5
   11ecc: 6118         	c.ld	a4, 0x0(a0)
   11ece: 8f3d         	c.xor	a4, a5
   11ed0: 28775713     	orc.b	a4, a4
   11ed4: 9f75         	c.not	a4
   11ed6: e705         	c.bnez	a4, 0x11efe <memchr+0xc0>
   11ed8: 8ebd         	c.xor	a3, a5
   11eda: 2876d713     	orc.b	a4, a3
   11ede: 9f75         	c.not	a4
   11ee0: e705         	c.bnez	a4, 0x11f08 <memchr+0xca>
```
It mistakenly uses `a5` for the XOR (the register that's being packed), and still uses `a3`. If I change the constraint to "=r":

```bash
   11ec4: 08b5f8b3     	packh	a7, a1, a1
   11ec8: 0918c8b3     	pack	a7, a7, a7
   11ecc: 6118         	c.ld	a4, 0x0(a0)
   11ece: 01174733     	xor	a4, a4, a7
   11ed2: 28775713     	orc.b	a4, a4
   11ed6: 9f75         	c.not	a4
   11ed8: e70d         	c.bnez	a4, 0x11f02 <memchr+0xc4>
   11eda: 0116c6b3     	xor	a3, a3, a7
   11ede: 2876d713      orc.b	a4, a3
   11ee2: 9f75         	c.not	a4
   11ee4: e705         	c.bnez	a4, 0x11f0c <memchr+0xce>

```

We're back to what the disassembly was in the first case. I agree it should be straightforward, but I'm not sure what I'm doing wrong here.

thebigclub:
I added the function to splat bytes. I'll look into `Zilsd`/`Zclsd` next.

thebigclub:
The array of two elements did not work and actually caused `clang` to explode. May end up filing a bug report. But I'm impressed at how clever compilers are. I have this:

```c
          uint64_t val;
          asm volatile ("ld %0, 0(%1)"
                        : "=R" (val)
                        : "r" (src)
          );
```

Need to use `ld` because `c.ld` has an error if it's not a compressed register, which the compiler does not pick up on (might be a bug, will dig into it). Later in the C code, I have `val >>= 32;` and the binary disassembles to:

```bash
   11ec2: 08b5f733     	packh	a4, a1, a1
   11ec6: 08e74733     	pack	a4, a4, a4
   11eca: 00053803     	ld	a6, 0x0(a0)
   11ece: 010747b3     	xor	a5, a4, a6
   11ed2: 2877d793      orc.b	a5, a5
   11ed6: 9ff5         	c.not	a5
   11ed8: e78d         	c.bnez	a5, 0x11f02 <memchr+0xc4>
   11eda: 011747b3      xor	a5, a4, a7
   11ede: 2877d793      orc.b	a5, a5
   11ee2: 9ff5         	c.not	a5
   11ee4: e785         	c.bnez	a5, 0x11f0c <memchr+0xce>
```

It's clever enough to use `a7` after the 32-bit shift.

Unfortunately, it does not work for `rv32e`. I get this when compiling:

```bash
   11fae: 08b5f733     	packh	a4, a1, a1
   11fb2: 08e74733     	pack	a4, a4, a4
   11fb6: 00053303     	ld	t1, 0x0(a0)
   11fba: 006747b3     	xor	a5, a4, t1
   11fbe: 2877d793     	orc.b	a5, a5
   11fc2: 9ff5          c.not	a5
   11fc4: e78d         	c.bnez	a5, 0x11fee <memchr+0xc4>
   11fc6: 007747b3     	xor	a5, a4, t2
   11fca: 2877d793     	orc.b	a5, a5
   11fce: 9ff5          c.not	a5
   11fd0: e785         	c.bnez	a5, 0x11ff8 <memchr+0xce>
```

It uses `t1` as the even register, which is obviously wrong. At least it got the shift right. Might also be another bug.

thebigclub:
Ignore my remarks about `t1` (I removed them). I thought it was an odd register, but I see it maps to `x6`, which is obviously even.

thebigclub:
Pretty sure `clang` has a bug when using any optimization level other than `-O0`:

```bash
$ cat << EOF > test.c
#include <stdint.h>

int fn(void *arg1, int arg2, int arg3, int arg4, int arg5) {
        uint64_t val;

        asm volatile ("c.ld %0, 0(%1)"
                      : "=R" (val)
                      : "r" (arg1)
        );

        return val + arg2 + arg3 + arg4 + arg5;
}
EOF
$ clang -march=rv32e_zclsd -O3 -c test.c
zclsd.c:6:16: error: invalid operand for instruction
    6 |         asm volatile ("c.ld %0, 0(%1)"
      |                       ^
<inline asm>:1:7: note: instantiated into assembly here
    1 |         c.ld t1, 0(a0)
      |              ^
1 error generated.
```

thebigclub:
Can run through all the combinations with:
```bash
for isa in i e; do for level in 0 1 2 3 s z; do (set -x; clang -march=rv32"$isa"_zclsd -O"$level" -c test.c); echo Return value: $?; done; done
```
I'll file a bug report tomorrow.

thebigclub:
Since I filed the bug report for the array case, I went ahead and filed the other one at the time as well.

thebigclub:
Turns out it's not a bug. You have to use `cR` and not just `R` in the inline assembly constraint for `Zclsd`.

thebigclub:
I've pushed a draft that uses `Zilsd` when available.

christian-herber-nxp:
> Turns out it's not a bug. You have to use `cR` and not just `R` in the inline assembly constraint for `Zclsd`. The array case was at least a bug.

I think you should not use c.ld, but just use ld. Then you leave it up to the compiler to allocate a compressed register and the assembler will generate c.ld encoding in that case.

thebigclub:
I did change it to `ld`, but it didn't dawn on me that the assembler would convert `ld` to `c.ld` if possible, which in retrospect makes sense, because that happens for all the other instructions that can be compressed.

Got rid of a macro just now that wasn't really needed. I might also use a `val2` variable for the first check in `memchr()`, to make it more clear that separate registers are being accessed for `Zilsd`. The comment in the second check mentions this, but I'm torn about making it more explicit.

christian-herber-nxp:
> I did change it to `ld`, but it didn't dawn on me that the assembler would convert `ld` to `c.ld` if possible, which in retrospect makes sense, because that happens for all the other instructions that can be compressed.

the compiler will only do the register allocation, and convert everything to assembly instructions. You can check the output by just compiling with -S. It is the assembler that picks up the assembly file and converts it to actual encodings.

newlib/libc/machine/riscv/memchr.c
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+  const unsigned char *end_addr = src + (length & ~(SZREG - 1));
+
+  if (src < end_addr)
+    {
+      uintxlen_t mask = __libc_splat_byte(d);
+
+#if __riscv_zilsd
+      uint64_t val;
```
christian-herber-nxp, line 89: it will be nicer to add a new type even, that is either uintxlen_t or uint64_t. Something like uintldstlen (load/store length)

newlib/libc/machine/riscv/memchr.c
```diff
@@ -0,0 +1,169 @@
...
+        {
+#if __riscv_zilsd
+          asm volatile ("ld %0, 0(%1)"
+                        : "=R" (val)
+                        : "r" (src)
+          );
+#else /* not riscv_zilsd */
+          val = (*(uintxlen_t*) src);
+#endif /* __riscv_zilsd */
+          val ^= mask;
```
christian-herber-nxp, line 104: for zilsd, this seems to xor uint32_t with uint64_t?

newlib/libc/machine/riscv/memchr.c
```diff
@@ -0,0 +1,169 @@
...
+              if (*src++ == d) return (void *) src - 1;
+              if (*src++ == d) return (void *) src - 1;
+              if (*src++ == d) return (void *) src - 1;
+#endif /* __riscv_xlen == 64 */
+              return (void *) src;
+#endif /* __riscv_zbb */
+            }
+#if __riscv_zilsd
+          /*
+            Even though the earlier value was XORed with the mask, the
```
christian-herber-nxp, line 128: i see. I think this is not a case of clever, but simply how the language works. 
mask is implicitly converted to uint64_t.
This is not clean, better to do it explicitly to avoid ambiguities.

thebigclub: What I think I'm going to do is use a `val2` in that section to make this more clear. I did it for the `memrchr()` case, and I'll do the same for `memchr()`.

newlib/libc/machine/riscv/memchr.c
thebigclub: I see, and then use that instead of having the macro check in `memchr()`/`memrchr()`. Good idea. I'll add it to `xlenint.h`.

thebigclub:
Added the new type and made the word access more explicit.

thebigclub:
Just pushed another update. Gave the variables better names.

thebigclub:
Fixed a bug with `memrchr()`.

thebigclub:
My `strlen()` patch was accepted, so pushed a rebase just now.

thebigclub:
I added the initial commit that copies the stock implementations.

thebigclub:
@christian-herber-nxp Unless you have more remarks, since the commit that just copied the stock implementations was the only change left, and that's been added to the patch series, I could send it to the mailing list.

PR #13, opened 2025-05-05
memmove speed optimized

m-fally:
Here is the current state of my optimization-suggestion for memmove().

I have used the common implementation of the function as a base, which first checks if at least one of the addresses is unaligned, and only uses an optimized algorithm if both addresses are aligned. Otherwhise it copies byte by byte.

I have added an unrolled loop for the destructive overlap case as well as increased the unroll-factor of the loop in the non-destructive overlap case. I chose an unroll factor of 9*SZREG since that is what is also used in memcpy(), and according to my tests it seems like a reasonable choice.

The other bigger change is aligning the source address when one of the addresses is unaligned. In that case, one word (or doubleword for rv64) is loaded at a time, and individual bytes are stored to the destination. For the vast majority of cases I get better or equally as good results when aligning the source than when aligning the destination.
This change also affects the case when source and destination are aligned, there is a destructive overlap and length is not a multiple of SZREG, because in this scenario the length is added to both source and destination, which causes them to become unaligned.

I have also tried always aligning the source and then checking whether source and destination have the same alignment. This performs better than checking for misalignment when at least one address is unaligned (especially when the addresses already have the same alignment). However, when the addresses are both aligned (or when -mno-strict-align is used) in a lot of cases this approach performs worse than just checking for misalignment of one of the addresses (in the case of -mno-strict-align performance is even slightly worse than in the original implementation).
Maybe it would be an idea to use the check for matching alignment only in the case of a destructive overlap, since having two aligned addresses and an "unaligned" length would produce the scenario where that approach performs well, and in the case of two aligned addresses and an aligned length it does not do too much worse than the other approach. But that would favour the case where length is not a multiple of SZREG, so I am not sure if it is good to do that.

The following tables compare the data for both approaches I have tried (checking for misalignment of at least one address and only aligning if one is misaligned vs. always aligning the source and then checking whether the alignments of source and destination match) and the original algorithm for rv32 and rv64:

[memmove_speed_opt_comparison_rv32.ods](https://github.com/user-attachments/files/20033361/memmove_speed_opt_comparison_rv32.ods)
[memmove_speed_opt_comparison_rv64.ods](https://github.com/user-attachments/files/20033362/memmove_speed_opt_comparison_rv64.ods)

With regards to incorporating zilsd I haven't had any success yet, since there are additional checks and alignment operations needed to be sure the addresses are aligned by 8 bytes.
I think I will try to also unroll the misaligned-copy-loop and see if in that case the extra overhead is worth it.


newlib/libc/machine/riscv/memmove-stub.c
christian-herber-nxp, line 1: the filename should no longer be memmove-stub.c but memmove.c

newlib/libc/machine/riscv/memmove-stub.c
```diff
@@ -1,14 +1,197 @@
...
-   of the FreeBSD License.   This program is distributed in the hope that
-   it will be useful, but WITHOUT ANY WARRANTY expressed or implied,
-   including the implied warranties of MERCHANTABILITY or FITNESS FOR
-   A PARTICULAR PURPOSE.  A copy of this license is available at
-   http://www.opensource.org/licenses.
-*/
 
-#if !defined(PREFER_SIZE_OVER_SPEED) && !defined(__OPTIMIZE_SIZE__)
-#include "../../string/memmove.c"
+static inline uint8_t __libc_aligned_xlen(void *dst, const void *src) {
```
christian-herber-nxp, line 11: name is not correct, as this can return 1 when not xlen aligned. better to call it check_fast_alignment or so.

newlib/libc/machine/riscv/memmove-stub.c
```diff
@@ -1,14 +1,197 @@
...
+#if defined(__riscv_misaligned_fast)
+	return 1;
+#else
+	return !(((uintxlen_t)src & (SZREG-1)) | ((uintxlen_t)dst & (SZREG-1)));
 #endif
+}
+
+
+/*SUPPRESS 20*/
```
christian-herber-nxp, line 20: ?

newlib/libc/machine/riscv/memmove-stub.c
```diff
@@ -1,14 +1,197 @@
...
+		src += length;
+		dst += length;
+
+		if (length >= SZREG) {
+
+			if (__libc_aligned_xlen(dst, src)) {
+				aligned_dst = (uintxlen_t*)dst;
+				aligned_src = (uintxlen_t*)src;
+
+				/* Copy 9X SZREG bytes at a time if possible.  */
```
christian-herber-nxp, line 41: that is obvious. the comment should state why 9 is your choice

newlib/libc/machine/riscv/memmove-stub.c
```diff
@@ -1,14 +1,197 @@
...
+	else { /* Case: Non-destructive overlap or no overlap at all. */
+
+		if (length >= SZREG) {
+
+			if (__libc_aligned_xlen(dst, src)) {
+				aligned_dst = (uintxlen_t*)dst;
+				aligned_src = (uintxlen_t*)src;
+
+				/* Copy 9X SZREG bytes at a time if possible.  */
+				while (length >= (SZREG*9)) {
```
christian-herber-nxp, line 122: when i see such relatively long code segments that i have seen before, I will usually ask you to make an inline function.
Makes the code more readable and maintainable. 
As a rule of thumb, if your function does not fit on one screen, you should probably structure it better

newlib/libc/machine/riscv/memmove-stub.c
```diff
@@ -1,14 +1,197 @@
-/* Copyright (c) 2019  SiFive Inc. All rights reserved.
```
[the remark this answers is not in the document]
m-fally: should I keep the Copyright notice from the stub file? i was wondering because it only had the include line, and the implementation in the common port does not have a copyright notice at all

newlib/libc/machine/riscv/memmove-stub.c
m-fally: i'm not sure i understand what you mean by "option to the compiler" – do you mean to have a variable that contains either -1 or 1 depending on the kind of overlap, so then the same code can be used for both scenarios?

christian-herber-nxp: > i'm not sure i understand what you mean by "option to the compiler" – do you mean to have a variable that contains either -1 or 1 depending on the kind of overlap, so then the same code can be used for both scenarios?

I formulated this weirdly but you got the point.

newlib/libc/machine/riscv/memmove-stub.c
m-fally: this comment comes from the stock implementation, I will look into why it's there

newlib/libc/machine/riscv/memmove-stub.c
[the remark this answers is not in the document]
m-fally: okay, so the SiFive license should stay?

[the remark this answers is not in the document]
christian-herber-nxp: yes just keep it.

newlib/libc/machine/riscv/memmove-stub.c
m-fally: I've been trying to figure out what this means, but didn't really find an answer. There are similiar comments in e.g. the `strcat()` and `strcpy()` stock implementations, but also without explanation. They have been there for a really long time also, earliest commit I could find is from 2000.
Maybe they could be directives for a static analysis tool, but I could not not find any tool that accepts this exact syntax/keyword.
I also looked at glibc and could not find any comments similiar to this in there either.

Should I maybe ask about this on the mailing list?

christian-herber-nxp: i agree it is likely for a static analysis tool. can be removed.

m-fally: ok, will do!

m-fally:
@christian-herber-nxp I've pushed an updated version of the implementation with the changes you suggested and tried to organize the data better/made some plots to visualize the differences between implementations:

For `rv32ic -mstrict-align`, there were 9 cases where the new version was slower than the original with a maximum difference of 10 retired instructions. For `rv64ic -mstrict-align` there was 1 case where the new version was slower with a difference of 2 retired instructions.
For `rv32ic -mno-strict-align` and `rv64ic -mno-strict-align`, the new version was faster or equally as fast as the original implementation in all test cases.
In the tables below, this new version is indicated as "Check for misalignment".

I have also tested a version based on the `memcpy()` algorithm, where I just added code to handle the destructive-overlap-case and made smaller adjustments so that some code can be reused in both overlap-scenarios (This implementation is indicated as "Check for matching alignment" in the table below).

All tests were run on `spike`. I have run tests with both source and destination address aligned, both addresses unaligned, source aligned and destination unaligned, and destination aligned and source unaligned.

Graphical plots for all scenarios tested can be found here: [https://cloud.servus.at/s/a7Gqnoigxak7y4q](https://cloud.servus.at/s/a7Gqnoigxak7y4q)

------------------------------------------------------------------------------------
**-march=rv32ic -mabi=ilp32 -mtune=thead-c906 -mstrict-align -O3**
|  | Number of Cases | Highest difference in instructions retired |
| ------------- | ------------- | ------------- |
|Check for misalignment was faster than check for matching alignment:|5152|-1541|
|Check for misalignment was slower than check for matching alignment:|3022|4321|
|Check for misalignment was equally as fast as check for matching alignment:|0||
||||
|Check for misalignment was faster than original:|8162|-4583|
|Check for misalignment was slower than original:|9|10|
|Check for misalignment was equally as fast as original:|3|
||||
|Check for matching alignment was faster than original:|8054|-8836|
|Check for matching alignment was slower than original:|115 |115|
|Check for matching alignment was equally as fast as original:|5||

Total number of tests: 8174

------------------------------------------------------------------------------------
**-march=rv32ic -mabi=ilp32 -mtune=thead-c906 -mno-strict-align -O3**
|  | Number of Cases | Highest difference in instructions retired |
| ------------- | ------------- | ------------- |
|Check for misalignment was faster than check for matching alignment:|8174|-176|
|Check for misalignment was slower than check for matching alignment:|0||
|Check for misalignment was equally as fast as check for matching alignment:|0|
||||
|Check for misalignment was faster than original:|8171|-1298|
|Check for misalignment was slower than original: |0||
|Check for misalignment was equally as fast as original:|3|
||||
|Check for matching alignment was faster than original:|7680|-1293|
|Check for matching alignment was slower than original:|492|142|
|Check for matching alignment was equally as fast as original:|2|

Total number of tests: 8174

------------------------------------------------------------------------------------
**-march=rv64ic -mabi=lp64 -mtune=thead-c906 -mstrict-align -O3**
|  | Number of Cases | Highest difference in instructions retired |
| ------------- | ------------- | ------------- |
|Check for misalignment was faster than check for matching alignment:|4712|-1801|
|Check for misalignment was slower than check for matching alignment:|3453|4185|
|Check for misalignment was equally as fast as check for matching alignment:|1|
||||
|Check for misalignment was faster than original:|8164|-5343|
|Check for misalignment was slower than original:|1|2|
|Check for misalignment was equally as fast as original:|1|
||||
|Check for matching alignment was faster than original:|7926|-9452|
|Check for matching alignment was slower than original:|239|275|
|Check for matching alignment was equally as fast as original:|1|

Total number of tests: 8166

------------------------------------------------------------------------------------
**-march=rv64ic -mabi=lp64 -mtune=thead-c906 -mno-strict-align -O3**
|  | Number of Cases | Highest difference in instructions retired |
| ------------- | ------------- | ------------- |
| Check for misalignment was faster than check for matching alignment: | 8166 | -360 |
| Check for misalignment was slower than check for matching alignment: | 0 |  |
| Check for misalignment was equally as fast as check for matching alignment: | 0 | |
||||
| Check for misalignment was faster than original: | 8151 | -654 |
| Check for misalignment was slower than original: | 0 |  |
| Check for misalignment was equally as fast as original: | 15 |
||||
| Check for matching alignment was faster than original:	| 6058 | -649 |
| Check for matching alignment was slower than original: | 2088 | 326 |
| Check for matching alignment was equally as fast as original: | 20 | |

Total number of tests: 8166

m-fally:
Since there are less cases where the "check for misalignment"-version is slower than the original `memmove()` implementation, I thought to stick with that one, but there are also cases where the "check for matching"-alignment version is a lot faster. So I was curious what you think?

newlib/libc/machine/riscv/memmove-stub.c
m-fally: just pushed a new suggestion for this

m-fally:
Forgot to add the information about code sizes, here it is:

**Original Implementation**
|| Code Size in Bytes |
|---|---|
|rv32ic -mstrict-align |326|
|rv32ic -mno-strict-align|300|
|rv64ic -mstrict-align |336|
|rv64ic -mno-strict-align|426|

**New Implementation (Check for misalignment)**
|| Code Size in Bytes |
|---|---|
|rv32ic -mstrict-align |1082|
|rv32ic -mno-strict-align|580|
|rv64ic -mstrict-align |1370|
|rv64ic -mno-strict-align|692|

**New Implementation (Check for matching alignment)**
|| Code Size in Bytes |
|---|---|
|rv32ic -mstrict-align |1082|
|rv32ic -mno-strict-align|614|
|rv64ic -mstrict-align |1284|
|rv64ic -mno-strict-align|676|

